PR #9, opened 2024-01-17
[ENH] Implemented query fields

### LINKED ISSUE #2: Implement query fields
- [x] query fields for the variables exposed by the API are added
- [x] query selections are reflected in the API request being sent out
- [x] UI element blocking (healthy control <-> diagnosis)
- [x] fetch the options for variables from the API at startup

rmanaem:
- Closes #2 

## Checklist

- [ ] Tests pass
- [ ] Checks pass

For new features:
- [ ] Tests have been added

For bug fixes:


src/App.tsx
```diff
@@ -1,111 +1,321 @@
...
+      renderInput={(params) => (
+        <TextField
+          // eslint-disable-next-line react/jsx-props-no-spreading
+          {...params}
+          InputLabelProps={params.InputLabelProps}
+          label={label}
+          className="w-full"
+        />
+      )}
+      onChange={(_, value) => onFieldChange(label, value?.id ?? null)}
```
surchs, line 43: aha, took me a second to get that `??` here means that if the left is either `undefined` or `null`, then the right will always be just `null`. good to know

src/App.css
surchs, line 1: do we even want to keep this file around? I think it could get a bit confusing if everything is styled either in mui or in tailwind, except for some very rare styles that live here

src/App.tsx
```diff
@@ -1,111 +1,321 @@
...
-function ResultContainer({result} : {result: Result[]}) {
+function ResultContainer({ result }: { result: Result[] }) {
   return (
-    <Stack spacing={2}>
-          {exampleResult.map((item) =>
-          <ResultCard key={item.dataset_uuid} nodeName={item.node_name} datasetName={item.dataset_name} datasetTotalSubjects={item.dataset_total_subjects} numMatchingSubjects={item.num_matching_subjects} imageModals={item.image_modals} />
-          )}
-    </Stack>
-  )
+    <div className="grid gap-4">
```
surchs, line 119: this, I have to say, is nice. Such a simple and clear way to describe the layout

src/App.tsx
```diff
@@ -1,111 +1,321 @@
...
+  apiQueryURL,
+  onSubmitQuery,
+}: {
+  diagnosisOptions: APIOption[];
+  assessmentOptions: APIOption[];
+  apiQueryURL: string;
+  onSubmitQuery: (url: string) => void;
+}) {
+  const [minAge, setMinAge] = useState<string | null>(null);
+  const [maxAge, setMaxAge] = useState<string | null>(null);
```
surchs, line 146: I think eventually we should probably treat these as numeric. OK to keep as strings for now, but I could imagine this getting confusing if we pass strings to a component that then needs to ensure that the "string" is not smaller than 0 and so on.

Maybe a TODO comment for now should make sure we remember to fix this in the final refactor

src/App.tsx
```diff
@@ -1,111 +1,321 @@
...
+  const [sex, setSex] = useState<string | null>(null);
+  const [diagnosis, setDiagnosis] = useState<string | null>(null);
+  const [isControl, setIsControl] = useState<boolean>(false);
+  const [minNumSessions, setMinNumSessions] = useState<string | null>(null);
+  const [assessmentTool, setAssessmentTool] = useState<string | null>(null);
+  const [imagingModality, setImagingModality] = useState<string | null>(null);;
```
surchs, line 152: just a note for later: since the modalities are fixed / not fetched, we could declare the valid options in the type hint. I believe it's

`type modalities = "T1w" | "T2w"` and so on. 

Again, more for later

src/App.tsx
```diff
@@ -1,111 +1,321 @@
...
+  const [result, setResult] = useState<Result[]>([]);
+
+  useEffect( () => {
+    async function fetchData(dataElementURI : string, setState : (options: APIOption[]) => void) {
+      try {
+        const response: AxiosResponse<RetrievedOption> = await axios.get(`${attributesURL}${dataElementURI}`);
+        if (response.data[dataElementURI].length === 0) {
+          // TODO: make into a toast
+        } else {
+          setState(response.data[dataElementURI].sort((a, b) => a.Label.localeCompare(b.Label)));
```
surchs, line 306: I think this would be a lot easier to read over if instead of setting state directly from inside the function we just returned the response and then handle the `setState` business outside - e.g. we wouldn't have these side-effect calls to just `fetchData(something)` that don't make clear what will happen after the fetch. We should also move the function out of the useEffect hook to keep it tight.

surchs (approved):
Very cool PR! Thanks @rmanaem, and congrats again on the rapid pace of this adoption / transition. Very excited to keep this going.

I have only one real comment (about the useEffect hook), the other ones are more notes. Take a look, and then it's :cook:

src/App.tsx
```diff
@@ -1,111 +1,321 @@
...
-    'http://purl.org/nidash/nidm#T1Weighted': {
-      name: 'T1',
-      style: 'bg-yellow-500 text-white',
-    },
-    'http://purl.org/nidash/nidm#T2Weighted': {
-      name: 'T2',
-      style: 'bg-green-600 text-white',
-    },
-  };
+const modalityOptions: object = modalities;
```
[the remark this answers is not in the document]
rmanaem: Good catch, we actually do the type hints in there so we can get rid of this.

src/App.tsx
rmanaem: This may be an issue cause the `TextField` itself returns a string even for type number but we can see if its possible to change that. I'll add a TODO.

src/App.tsx
rmanaem: > I think this would be a lot easier to read over if instead of setting state directly from inside the function we just returned the response and then handle the setState business outside

Since its an `async` function it returns a promise which requires another `async-wait` to set the state since the `seState` functions don't allow for promises.

> We should also move the function out of the useEffect hook to keep it tight.

Tried that as we discussed but Eslint didn't seem to like it. Since the `fetch` function wasn't defined inside of the hook it asked to either add it to the dependency array which would no longer be the equivalent of `componentDidMount` or remove it. A workaround suggested by the GPT was to use [`useCallBack`](https://react.dev/reference/react/useCallback) and put it in the dependency array of that and then use it in the `useEffect` which upon searching a bit seemed like an overkill and didn't seems appropriate use of `useCallback`.

surchs:

PR #16, opened 2024-02-05
[ENH] Implemented UI improvements

### LINKED ISSUE #14: UI improvements
- [x] Alerts
  - [x] Navbar
  - [x] Logo
  - [x] favicon
  - [x] Form validation
  - ~[ ] name ellipses~ No need for ellipses expansion and tooltip since dataset names seems to be wrapping around nicely
  - [x] summary stats
  - [x] toasts
  - [x] Button loading
  - [x] non-white background color of the tool group buttons on hover

rmanaem:
- Closes #14 

## Checklist

- [ ] Tests pass
- [ ] Checks pass

For new features:
- [ ] Tests have been added

For bug fixes:


src/components/QueryForm.tsx
```diff
@@ -1,179 +1,97 @@
...
-        setAssessmentTool(value);
-        break;
-      case 'Imaging modality':
-        setImagingModality(value);
-        break;
-      default:
-        break;
-    }
-  }
+  const minAgeBiggerThanMax: boolean = minAge && maxAge ? minAge > maxAge : false;
```
surchs, line 57: neato!

surchs: how about `minAgeExceedsMaxAge` or adding `Error` in there to drive home that this thing being `true` is a bad thing?

src/components/QueryForm.tsx
```diff
@@ -1,179 +1,97 @@
...
-  const [minNumSessions, setMinNumSessions] = useState<string | null>(null);
-  const [assessmentTool, setAssessmentTool] = useState<FieldInput>(null);
-  const [imagingModality, setImagingModality] = useState<FieldInput>(null);
-  const [searchParams, setSearchParams] = useSearchParams();
+  const [minAgeError, setMinAgeError] = useState(false);
+  const [minAgeHelperText, setMinAgeHelperText] = useState('');
+  const [maxAgeError, setMaxAgeError] = useState(false);
+  const [maxAgeHelperText, setMaxAgeHelperText] = useState('');
+  const [minNumSessionsError, setMinNumSessionsError] = useState(false);
+  const [minNumSessionsHelperText, setMinNumSessionsHelperText] = useState('');
```
surchs, line 55: I'm not sure why any of these are stateful. All of these can be computed directly from the props - and the props are handled by the parent now (i.e. App.tsx). So unless I'm missing something, the `validate` function can go, and the only eventhandler here is the forwarded `updateCategoricalOption` or `updateContinuousOption` handlers from the parent

src/components/QueryForm.tsx
```diff
@@ -1,179 +1,97 @@
...
-    if (Array.isArray(value)) {
-      value.forEach((v) => {
-        queryParamObject.append(param, v.id);
-      });
+    if (numberValue < min) {
+      setError(true);
+      setHelperText(`Value must be greater than or equal to ${min}`);
+    } else if (max && numberValue > max) {
+      setError(true);
+      setHelperText(`Value must be less than or equal to ${max}`);
```
surchs, line 81: as mentioned before, I think this whole function can go and the stateful error handling replaced by plain JS. This will also help with the slightly funky React.Dispatch type hints that are needed here

src/components/QueryForm.tsx
```diff
@@ -235,21 +168,25 @@ function QueryForm({
...
       </div>
-      <div className={isFederationAPI ? 'col-span-2 row-start-5' : 'col-span-2 row-start-4'}>
+      <div className={isFederationAPI ? 'col-span-2 row-start-6' : 'col-span-2 row-start-5'}>
         <ContinuousField
+          error={minNumSessionsError}
+          helperText={minNumSessionsHelperText}
           label="Minimum number of sessions"
-          onFieldChange={(label, value) => updateContinuousQueryParams(label, value)}
+          onFieldChange={(label, value) =>
+            validate(label, value, setMinNumSessionsError, setMinNumSessionsHelperText, 0)
```
surchs, line 177: as mentioned above, I think we can just do the same here as we do for categorical variables, i.e. just pass the `updateContinuousOptions` handler (or what it is called) and handle the validation in plain JS based on the props

src/components/QueryForm.tsx
```diff
@@ -260,11 +197,18 @@ function QueryForm({
...
         <Button
+          disabled={disableSubmit}
           variant="contained"
-          endIcon={<SendIcon />}
-          onClick={() => onSubmitQuery(constructQueryURL())}
+          endIcon={
+            loading ? (
+              <CircularProgress size="20px" thickness={5.5} className="text-white" />
+            ) : (
+              <SendIcon />
```
surchs, line 208: neat!

src/components/ResultContainer.tsx
```diff
@@ -174,6 +184,9 @@ function ResultContainer({ result }: { result: Result[] | null }) {
             }
           />
         </div>
+        <div className="col-end-5 justify-self-end">
+          <Typography variant="body1">{summaryStats()}</Typography>
```
surchs, line 188: why does this have to be a function call?

src/components/ResultContainer.tsx
```diff
@@ -12,7 +12,17 @@ function ResultContainer({ result }: { result: Result[] | null }) {
...
+    let datasets = 0;
+    let subjects = 0;
+    if (result) {
+      result.forEach((item) => {
+        datasets += 1;
+        subjects += item.num_matching_subjects;
+      });
+    }
+    return `Summary stats: ${datasets} datasets, ${subjects} subjects`;
+  }
```
surchs, line 26: This code will run every time the component renders anyway, so let's lift this out of the function and put it directly into the plain JS part to make that clear!

surchs: i.e. have a `const summaryText = ...` variable that contains all this text

src/components/SnackStack.tsx
```diff
@@ -0,0 +1,33 @@
...
+import SnackbarToast from './SnackToast';
+
+function SnackStack() {
+  const { toastsPack } = useSnackStack();
+
+  const firstToast = toastsPack[0];
+
+  return (
+    <Snackbar
+      open={!!firstToast}
```
surchs, line 12: is this a typo? Looks like double negative?

src/components/SnackStack.tsx
```diff
@@ -0,0 +1,33 @@
...
+      autoHideDuration={null}
```
surchs, line 13: `null` seems to be the default. Are you setting it on purpose? else I'd just remove

src/App.tsx
```diff
@@ -1,9 +1,22 @@
...
+  FieldInput,
+  FieldInputOption,
+  Result,
+} from './utils/types';
 import QueryForm from './components/QueryForm';
 import ResultContainer from './components/ResultContainer';
+import Navbar from './components/Navbar';
+import { useSnackStack } from './components/SnackStackProvider';
+import SnackStack from './components/SnackStack';
 import './App.css';
```
surchs, line 20: Commenting here: This is a really impressive amount of work to get the stacking snackbar to work. But now that's our code to keep alive and I would very much avoid that. Essentially

- MUI / Android says: do not stack toasts
- but we want to do it anyway, breaking MUI / Android visual spec
- MUI says: good luck, you're on your own. But you should know that if you do this, here is a third party library that handles it for you: https://mui.com/material-ui/react-snackbar/#third-party-integrations
- we say: thanks a lot, but we'll both break with MUI and not use the third-party app

That last point is where I disagree. I think we should either

- stop stacking toasts (not good because toasts are nice) OR
- just use notistack: https://notistack.com/features/basic and don't bother adding all this stuff ourselves

surchs (changes requested):
Really impressive amount of work @rmanaem! Thanks a lot for that :tada: - and sorry for the long wait for the review. Everything looks very nice, and does what it is supposed to - we'll get to the bottom of the performance issues in a future PR.

That said, I have two main things I would address:
- stateful error handling in the QueryForm component. I think we should not keep state there and instead just compute errors based on props directly (you're doing that for some, but not all errors). See comments
- Snackbar: I think this is a beautiful piece of code and quite a complex achievement. But as I commented inline in the App, there is an existing solution for it and by not using that we now have to maintain a complex piece of code. So I'd say: let's adopt the third party solution instead.

Let me know if you'd like me to push these changes direct to the branch. I understand that we'll have to likely fix #19 then - happy to help with that too.

Some minor comments throughout the PR, take a look

src/App.tsx
```diff
@@ -12,8 +25,26 @@ function App() {
...
     { NodeName: 'All', ApiURL: 'allNodes' },
   ]);
+
+  const [alertDismissed, setAlertDismissed] = useState<boolean>(false);
+
+  const [loading, setLoading] = useState<boolean>(false);
+
   const [result, setResult] = useState<Result[] | null>(null);
 
+  const [node, setNode] = useState<FieldInput>([{ label: 'All', id: 'allNodes' }]);
```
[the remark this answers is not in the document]
rmanaem: To help with readability they belong to different sections of the code.

src/App.tsx
```diff
@@ -1,9 +1,22 @@
...
+  NodeOption,
+  FieldInput,
+  FieldInputOption,
+  Result,
+} from './utils/types';
 import QueryForm from './components/QueryForm';
 import ResultContainer from './components/ResultContainer';
+import Navbar from './components/Navbar';
+import { useSnackStack } from './components/SnackStackProvider';
+import SnackStack from './components/SnackStack';
```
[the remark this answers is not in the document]
rmanaem: I think prettier (in the way that we have it set up now) does some import ordering but perhaps not to the extent that we'd want. I opened #21 to address this in another PR.

src/App.tsx
```diff
@@ -24,13 +55,20 @@ function App() {
           `${attributesURL}${dataElementURI}`
         );
         if (response.data[dataElementURI].length === 0) {
-          // TODO: make into a toast
+          addToast({
+            key: uuidv4(),
+            message: `No options found for ${dataElementURI.slice(3)}`,
+            severity: 'info',
+          });
         } else {
```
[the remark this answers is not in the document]
rmanaem: If I recall correctly we tried refactoring that but since the function is asynchronous and returns a promise Typescript/linter doesn't like assigning it directly to a variable and requires an anonymous function that implementation leads to verbosity in addition, if the `fetchOption` function is used inside the useEffect it is required to either be defined inside it as it currently is or be part of the dependency array which would lead to the useEffect running when we don't necessarily want it to.
That being said, I do agree that the function needs to be refactored to handle less responsibility.

src/App.tsx
```diff
@@ -50,36 +91,237 @@ function App() {
 
     fetchOptions('nb:Diagnosis', setDiagnosisOptions);
     fetchOptions('nb:Assessment', setAssessmentOptions);
-  }, []);
+  }, []); // eslint-disable-line react-hooks/exhaustive-deps
```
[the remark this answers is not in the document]
rmanaem: I spent some time looking into this when I came across it. Unfortunately having `addToast` as a dependency leads to the useEffect running over and over which in terms of functionality breaks the toast. Essentially you'll end up with a bunch of toasts being spammed on the side of the screen.
To ensure we wouldn't run into an old state being used I memoized the `addToast` and the `contextValue` in `SnackStackProvider`.

src/App.tsx
```diff
@@ -50,36 +91,237 @@ function App() {
...
+          setNode(matchedOptions);
+        }
+      }
+    }
+  }, [searchParams, setSearchParams, nodeOptions, node]);
 
-  async function submitQuery(url: string) {
+  function showAlert() {
+    if (node && Array.isArray(node)) {
+      const openNeurIsAnoOption = nodeOptions.find((n) => n.NodeName === 'OpenNeuro');
```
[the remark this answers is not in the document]
rmanaem: Nice catch!

src/components/SnackStack.tsx
rmanaem: It's one way to cast to boolean https://stackoverflow.com/questions/10467475/double-negation-in-javascript-what-is-the-purpose

src/App.tsx
```diff
@@ -50,36 +91,237 @@ function App() {
...
+   * @remarks
+   * This is a utility function to used to help construct the query URL using a URLSearchParams object.
+   *
+   * @param param - The name of the query parameter
+   * @param value - The value of the query parameter
+   * @param queryParamObject - The query parameter object which contains the query parameters
+   * @returns void
+   */
+  function setQueryParam(param: string, value: FieldInput, queryParamObject: URLSearchParams) {
+    if (Array.isArray(value)) {
```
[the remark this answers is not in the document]
rmanaem: Yea TypeScript is rather strict about that.

src/App.tsx
```diff
@@ -50,36 +91,237 @@ function App() {
...
+   * This function utilizes the `setQueryParam` function to set categorical query parameters.
+   *
+   * @returns The query URL.
+   */
+  function constructQueryURL() {
+    const queryParams = new URLSearchParams();
+
+    setQueryParam('node_url', node, queryParams);
+    queryParams.set('min_age', minAge ? minAge.toString() : '');
+    queryParams.set('max_age', maxAge ? maxAge.toString() : '');
```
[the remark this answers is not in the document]
surchs: very good!

src/components/ContinuousField.tsx
```diff
@@ -1,21 +1,27 @@
...
+  helperText?: string;
   label: string;
   onFieldChange: (fieldLabel: string, value: string) => void;
-}) {
+}
+
+function ContinuousField({ error, helperText, label, onFieldChange }: ContinuousFieldProps) {
   return (
-    // TODO: see if we can make it so TextField returns type number instead of string as its doing now
     <TextField
```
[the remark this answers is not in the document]
rmanaem: This is one of the things I came to after a couple of iterations but I think the main reason was it would allow text to be typed but it wouldn't handle it properly once the element is out of focus.

Peek 2024-02-13 09-32

src/components/Navbar.tsx
```diff
@@ -0,0 +1,54 @@
+import { useState, useEffect } from 'react';
+import { Toolbar, Typography, IconButton, Badge } from '@mui/material';
+import GitHubIcon from '@mui/icons-material/GitHub';
+
+function Navbar() {
+  const [latestReleaseTag, setLatestReleaseTag] = useState('');
+
+  useEffect(() => {
+    async function fetchLatestRelease() {
```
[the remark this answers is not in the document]
rmanaem: I believe this was due to the asynchronous nature of the requests being made. They needed to been enclosed in an async function. Will refactor so the anonymous function is async.

src/components/QueryForm.tsx
```diff
@@ -1,179 +1,97 @@
...
+  diagnosis: FieldInput;
+  isControl: boolean;
+  setIsControl: (value: boolean) => void;
+  assessmentTool: FieldInput;
+  imagingModality: FieldInput;
+  updateCategoricalQueryParams: (label: string, value: FieldInput) => void;
+  updateContinuousQueryParams: (label: string, value: number | null) => void;
+  loading: boolean;
+  onSubmitQuery: () => void;
 }) {
```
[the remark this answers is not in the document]
rmanaem: I don't think so. My understanding is that we need to define an interface/type for components (e.g., ContinuousField) when they have an optional prop.

src/App.tsx
rmanaem: I agree let's use notistack.

src/App.tsx
```diff
@@ -12,74 +23,302 @@ function App() {
...
+  const [assessmentTool, setAssessmentTool] = useState<FieldInput>(null);
+  const [imagingModality, setImagingModality] = useState<FieldInput>(null);
+  const [searchParams, setSearchParams] = useSearchParams();
+
   useEffect(() => {
-    async function fetchOptions(
-      dataElementURI: string,
-      setOptions: (options: AttributeOption[]) => void
-    ) {
+    async function tryThisOptionsGetter(dataElementURI: string) {
```
surchs, line 45: as much as I like the function name, maybe we should refactor this to something more typical. Maybe `getAttributes()`?

src/components/Navbar.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+  useEffect(() => {
+      // TODO: replace with react-query-tool once there is a release
+      const GHApiURL = 'https://api.github.com/repos/neurobagel/query-tool/releases/latest';
+      axios.get(GHApiURL)
+      .then((response) => {
+        const {data} = response;
+        setLatestReleaseTag(data.tag_name);
+      }).catch(() => {
+        setLatestReleaseTag('beta');
```
surchs, line 17: neat!

surchs (approved):
very nice :tada: ! Thanks @rmanaem.

One refactor suggestion, then :cook:

src/App.tsx
rmanaem: oh yea forgot about that.

surchs:

PR #25, opened 2024-02-15
[FIX] Refactored syncing of the `URLSearchParams` and selected nodes

### LINKED ISSUE #24: Refactor the synching of the `URLSearchParams` and selected nodes

rmanaem:
- Closes #24 

## Checklist

- [x] Tests pass
- [x] Checks pass

For new features:
- [ ] Tests have been added

For bug fixes:


src/App.tsx
```diff
@@ -41,6 +40,10 @@ function App() {
   const [imagingModality, setImagingModality] = useState<FieldInput>(null);
   const [searchParams, setSearchParams] = useSearchParams();
 
+  const node: FieldInputOption[] = nodeOptions
+    .filter((n) => searchParams.getAll('node').includes(n.NodeName))
+    .map((n) => ({ label: n.NodeName, id: n.ApiURL }));
```
surchs, line 45: maybe use a different name here, so we don't have `n` twice - that's confusing

surchs: how about `availableNode` and `selectedNode`?

src/App.tsx
```diff
@@ -98,42 +101,29 @@ function App() {
...
-          );
-          setNode(filteredNode);
-          setSearchParams({ node: filteredNode.map((n) => n.label) });
-        } else if (Array.isArray(node) && !arraysEqual(node, matchedOptions)) {
-          setNode(matchedOptions);
+          const filteredNodeNames = matchedNodeNames.filter((nodeName) => nodeName !== 'All');
+          setSearchParams({ node: filteredNodeNames });
         }
-      }
+      }
```
surchs, line 124: extra whitespace seems superfluous. mind running prettier again?

surchs (approved):
nice! thanks @rmanaem

one suggestion and then :cook:

src/App.tsx
rmanaem: > maybe use a different name here, so we don't have n twice - that's confusing

Shouldn't be the same since we're chaining of off the available nodes? We filter the options and then map the same filtered options.

surchs: the same type/shape, yes. But it's a different variable name. The only thing that separates them is the scope - but it makes it tricky to read / possible to get confused

surchs:

PR #31, opened 2024-02-21
[MNT] Implemented component tests

### LINKED ISSUE #20: component tests

rmanaem:
- Closes #20 

## Checklist

- [x] Tests pass
- [x] Checks pass

For new features:
- [x] Tests have been added

For bug fixes:

### DIFF
--- a/.github/workflows/component-test.yaml
+++ b/.github/workflows/component-test.yaml
@@ -1,0 +1,34 @@
+name: component tests
+
+on:
+  push:
+    branches:
+      - maint-20
+  pull_request:
+  workflow_dispatch:
+
+jobs:
+  component-test:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Set up node env
+        uses: actions/setup-node@v4.0.1
+        with:
+          node-version: 20
+
+      - name: Create .env file
+        run: |
+          echo -e "NB_API_QUERY_URL=${{ vars.NB_API_QUERY_URL }}\nNB_IS_FEDERATION_API=${{ vars.NB_IS_FEDERATION_API }}" > .env
+
+      - name: build
+        run: npm install && npm run build
+
+      - name: Run end to end tests
+        uses: cypress-io/github-action@v6
+        with:
+          wait-on: http://localhost:5173
+            start: npm run preview
+          component: true
--- a/.github/workflows/component-test.yaml
+++ b/.github/workflows/component-test.yaml
@@ -14,11 +14,6 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v4
 
-      - name: Set up node env
-        uses: actions/setup-node@v4.0.1
-        with:
-          node-version: 20
-
       - name: Create .env file
         run: |
           echo -e "NB_API_QUERY_URL=${{ vars.NB_API_QUERY_URL }}\nNB_IS_FEDERATION_API=${{ vars.NB_IS_FEDERATION_API }}" > .env
@@ -26,9 +21,7 @@ jobs:
       - name: build
         run: npm install && npm run build
 
-      - name: Run end to end tests
+      - name: Run component tests
         uses: cypress-io/github-action@v6
         with:
-          wait-on: http://localhost:5173
-          start: npm run preview
           component: true
--- a/.github/workflows/component-test.yaml
+++ b/.github/workflows/component-test.yaml
@@ -3,7 +3,7 @@ name: component tests
 on:
   push:
     branches:
-      - maint-20
+      - main
   pull_request:
   workflow_dispatch:
 
--- a/.github/workflows/component-test.yaml
+++ b/.github/workflows/component-test.yaml
@@ -30,5 +30,5 @@ jobs:
         uses: cypress-io/github-action@v6
         with:
           wait-on: http://localhost:5173
-            start: npm run preview
+          start: npm run preview
           component: true
--- a/cypress/component/CategoricalField.cy.tsx
+++ b/cypress/component/CategoricalField.cy.tsx
@@ -53,17 +53,17 @@ describe('CategoricalField', () => {
     );
   });
   it('Fires onFieldChange event handler with the appropriate payload when a value is selected', () => {
-    const onFieldChange = cy.spy().as('onFieldChange');
+    const onFieldChangeSpy = cy.spy().as('onFieldChangeSpy');
     cy.mount(
       <CategoricalField
         label={props.label}
         options={props.options}
-        onFieldChange={onFieldChange}
+        onFieldChange={onFieldChangeSpy}
         inputValue={props.inputValue}
       />
     );
     cy.get('[data-cy="Categorical Field-categorical-field"]').type('Option 1{downarrow}{enter}');
-    cy.get('@onFieldChange').should('have.been.calledWith', 'Categorical Field', {
+    cy.get('@onFieldChangeSpy').should('have.been.calledWith', 'Categorical Field', {
       id: '1',
       label: 'Option 1',
     });
--- a/cypress/component/CategoricalField.cy.tsx
+++ b/cypress/component/CategoricalField.cy.tsx
@@ -1,0 +1,38 @@
+import CategoricalField from "../../src/components/CategoricalField";
+
+const props = {
+    label: "Categorical Field",
+    options: [
+        { id: "1", label: "Option 1" },
+        { id: "2", label: "Option 2" },
+        { id: "3", label: "Option 3" },
+    ],
+    onFieldChange: () => {},
+    multiple: false,
+    inputValue: null,
+};
+
+describe("CategoricalField", () => {
+    it("Displays a MUI Autocomplete with the label and options passed as props", () => {
+        cy.mount(<CategoricalField label={props.label} options={props.options} onFieldChange={props.onFieldChange} inputValue={props.inputValue} />);
+        cy.get('[data-cy="Categorical Field-categorical-field"]').should("be.visible")
+        cy.get('[data-cy="Categorical Field-categorical-field"] label').should("contain", "Categorical Field")
+        props.options.forEach((option) => {
+            cy.get('[data-cy="Categorical Field-categorical-field"]').type(`${option.label}{downarrow}{enter}`)
+            cy.get('[data-cy="Categorical Field-categorical-field"] input').should("have.value", option.label);
+            cy.get('[data-cy="Categorical Field-categorical-field"]').clear();
+            
+        })
+    })
+    it("Displays the input value passed as props", () => {
+        cy.mount(<CategoricalField label={props.label} options={props.options} onFieldChange={props.onFieldChange} inputValue={props.options[0]} />);
+        cy.get('[data-cy="Categorical Field-categorical-field"] input').should("have.value", "Option 1")
+
+    })
+    it("Fires onFieldChange event handler with the appropriate payload when a value is selected", () => {
+        const onFieldChange = cy.spy().as("onFieldChange");
+        cy.mount(<CategoricalField label={props.label} options={props.options} onFieldChange={onFieldChange} inputValue={props.inputValue} />);
+        cy.get('[data-cy="Categorical Field-categorical-field"]').type("Option 1{downarrow}{enter}")
+        cy.get('@onFieldChange').should("have.been.calledWith", "Categorical Field", {id: "1", label: "Option 1"});
+    })
+})
--- a/cypress/component/CategoricalField.cy.tsx
+++ b/cypress/component/CategoricalField.cy.tsx
@@ -1,38 +1,71 @@
-import CategoricalField from "../../src/components/CategoricalField";
+import CategoricalField from '../../src/components/CategoricalField';
 
 const props = {
-    label: "Categorical Field",
-    options: [
-        { id: "1", label: "Option 1" },
-        { id: "2", label: "Option 2" },
-        { id: "3", label: "Option 3" },
-    ],
-    onFieldChange: () => {},
-    multiple: false,
-    inputValue: null,
+  label: 'Categorical Field',
+  options: [
+    { id: '1', label: 'Option 1' },
+    { id: '2', label: 'Option 2' },
+    { id: '3', label: 'Option 3' },
+  ],
+  onFieldChange: () => {},
+  multiple: false,
+  inputValue: null,
 };
 
-describe("CategoricalField", () => {
-    it("Displays a MUI Autocomplete with the label and options passed as props", () => {
-        cy.mount(<CategoricalField label={props.label} options={props.options} onFieldChange={props.onFieldChange} inputValue={props.inputValue} />);
-        cy.get('[data-cy="Categorical Field-categorical-field"]').should("be.visible")
-        cy.get('[data-cy="Categorical Field-categorical-field"] label').should("contain", "Categorical Field")
-        props.options.forEach((option) => {
-            cy.get('[data-cy="Categorical Field-categorical-field"]').type(`${option.label}{downarrow}{enter}`)
-            cy.get('[data-cy="Categorical Field-categorical-field"] input').should("have.value", option.label);
-            cy.get('[data-cy="Categorical Field-categorical-field"]').clear();
-            
-        })
-    })
-    it("Displays the input value passed as props", () => {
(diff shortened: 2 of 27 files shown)


.github/workflows/component-test.yaml
```diff
@@ -0,0 +1,27 @@
...
+jobs:
+  component-test:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Create .env file
+        run: |
+          echo -e "NB_API_QUERY_URL=${{ vars.NB_API_QUERY_URL }}\nNB_IS_FEDERATION_API=${{ vars.NB_IS_FEDERATION_API }}" > .env
```
surchs, line 19: I don't think we need that part for the component test, is that right? If yes I'd maybe add a comment to say so above this line and explain that we add it anyway to stay consistent to the regular deployment

cypress/component/CategoricalField.cy.tsx
```diff
@@ -0,0 +1,71 @@
...
+        inputValue={props.options[0]}
+      />
+    );
+    cy.get('[data-cy="Categorical Field-categorical-field"] input').should(
+      'have.value',
+      'Option 1'
+    );
+  });
+  it('Fires onFieldChange event handler with the appropriate payload when a value is selected', () => {
+    const onFieldChange = cy.spy().as('onFieldChange');
```
surchs: :tada: great find!
```suggestion
    const onFieldChangeSpy = cy.spy().as('onFieldChange');
```
How about that - or something to differentiate from the fixtures you are defining above.

surchs: More generally speaking - reusing the same variable name over and over again in different context can quickly lead to confusion: "which is it now, onFieldChange the fixture, onFieldChange the prop, onFieldChange the spy, onFieldChange the alias?" Try differentiating where that is reasonably possible to make more readable

cypress/component/ContinuousField.cy.tsx
```diff
@@ -0,0 +1,42 @@
...
+    cy.get('[data-cy="Continuous Field-continuous-field"] label').should(
+      'contain',
+      'Continuous Field'
+    );
+    cy.get('[data-cy="Continuous Field-continuous-field"] label').should(
+      'not.have.class',
+      'Mui-error'
+    );
+  });
+  it('Displays a MUI Textfield in error state with the helper text passed as a prop', () => {
```
surchs:
```suggestion
  it('Displays a MUI Textfield in error state when the helper text is not empty', () => {
```

cypress/component/ContinuousField.cy.tsx
```diff
@@ -0,0 +1,42 @@
+import ContinuousField from '../../src/components/ContinuousField';
+
+const props = {
+  helperText: 'This is a helper text',
+  label: 'Continuous Field',
+  onFieldChange: () => {},
+};
+
+describe('ContinuousField', () => {
+  it('Displays a MUI Textfield with the label passed as a prop', () => {
```
surchs:
```suggestion
  it('Displays a MUI Textfield and a label as passed via prop', () => {
```

cypress/component/ContinuousField.cy.tsx
```diff
@@ -0,0 +1,42 @@
...
+      />
+    );
+    cy.get('[data-cy="Continuous Field-continuous-field"] p').should(
+      'contain',
+      'This is a helper text'
+    );
+    cy.get('[data-cy="Continuous Field-continuous-field"] label').should('have.class', 'Mui-error');
+  });
+  it('Fires onFieldChange event handler with the appropriate payload when a value is entered', () => {
+    const onFieldChange = cy.spy().as('onFieldChange');
```
surchs: same as above with onFieldChange

cypress/component/DownloadResultButton.cy.tsx
```diff
@@ -0,0 +1,45 @@
...
+        identifier={props.identifier}
+        disabled
+        handleClick={props.handleClick}
+      />
+    );
+    cy.get('[data-cy="test-download-results-button"]').trigger('mouseover', { force: true });
+    cy.get('.MuiTooltip-tooltip').should('contain', 'Please select at least one dataset');
+  });
+  it('Fires the handleClick event handler with the appropriate payload when the button is clicked', () => {
+    const handleClick = cy.spy().as('handleClick');
```
surchs: :point_up_2:

cypress/component/NBDialog.cy.tsx
```diff
@@ -0,0 +1,28 @@
...
+      'contain',
+      'The command for automatically getting the data currently only applies to datasets available through datalad.'
+    );
+  });
+  it("Doesn't display the dialog when open prop is set to false", () => {
+    cy.mount(<NBDialog open={false} onClose={props.onClose} />);
+    cy.get('[data-cy="nb-dialog"]').should('not.exist');
+  });
+  it('Fires onClose event handler when the close button is clicked', () => {
+    const onClose = cy.spy().as('onClose');
```
surchs: :point_up_2:

cypress/component/Navbar.cy.tsx
```diff
@@ -0,0 +1,25 @@
+import Navbar from '../../src/components/Navbar';
+
+describe('Navbar', () => {
+  it('Displays a MUI Toolbar with logo, title, subtitle, documentation link, and GitHub link', () => {
+    cy.mount(<Navbar />);
+    cy.get("[data-cy='navbar']").should('be.visible');
+    cy.get("[data-cy='navbar'] img").should(
+      'have.attr',
+      'src',
+      'https://raw.githubusercontent.com/neurobagel/documentation/main/docs/imgs/logo/neurobagel_logo.png'
```
surchs: this is very cool as a way to test the attribute. But would we really want the test to fail just because the logo now has a different URL? We could just make sure that an image is loaded in this space. And if we really want to make sure that the right logo is there, cypress has plugins for visual testing: https://docs.cypress.io/guides/tooling/visual-testing#Open-source - but maybe for later

cypress/component/QueryForm.cy.tsx
```diff
@@ -0,0 +1,167 @@
...
+        updateContinuousQueryParams={updateContinuousQueryParams}
+        loading={props.loading}
+        onSubmitQuery={props.onSubmitQuery}
+      />
+    );
+    cy.get('[data-cy="Minimum age-continuous-field"]').type('10');
+    cy.get('@updateContinuousQueryParams').should('have.been.calledWith', 'Minimum age', 10);
+  });
+  it('Fires the onSubmitQuery event handler when the submit button is clicked', () => {
+    const onSubmitQuery = cy.spy().as('onSubmitQuery');
```
surchs: :point_up_2:

cypress/component/Navbar.cy.tsx
surchs: More generally: I think it'd be better to store the logo inside the app / repo. It's not like we change this often - and as it is, the logo is a network request itself. I realize this is a change for the navbar component - commenting here for simplicity

cypress/component/QueryForm.cy.tsx
```diff
@@ -0,0 +1,167 @@
...
+        diagnosis={props.diagnosis}
+        isControl={props.isControl}
+        minNumSessions={props.minNumSessions}
+        setIsControl={props.setIsControl}
+        assessmentTool={props.assessmentTool}
+        imagingModality={props.imagingModality}
+        updateCategoricalQueryParams={props.updateCategoricalQueryParams}
+        updateContinuousQueryParams={props.updateContinuousQueryParams}
+        loading={props.loading}
+        onSubmitQuery={props.onSubmitQuery}
```
surchs, line 70: This may be a misunderstanding, but to assert that these fields show up at all, why do you need to provide every prop? Are they all required / do not have defaults?

cypress/component/QueryForm.cy.tsx
```diff
@@ -0,0 +1,167 @@
...
+    cy.get('[data-cy="Sex-categorical-field"]').should('be.visible');
+    cy.get('[data-cy="Diagnosis-categorical-field"]').should('be.visible');
+    cy.get('[data-cy="healthy-control-checkbox"]').should('be.visible');
+    cy.get('[data-cy="Minimum number of sessions-continuous-field"]').should('be.visible');
+    cy.get('[data-cy="Assessment tool-categorical-field"]').should('be.visible');
+    cy.get('[data-cy="Imaging modality-categorical-field"]').should('be.visible');
+    cy.get('[data-cy="submit-query-button"]').should('be.visible');
+  });
+  it('Fires updateCategoricalQueryParams event handler with the appropriate payload when a categorical field is selected', () => {
+    const updateCategoricalQueryParams = cy.spy().as('updateCategoricalQueryParams');
```
surchs: :point_up_2:

cypress/component/QueryForm.cy.tsx
```diff
@@ -0,0 +1,167 @@
...
+    );
+
+    cy.get('[data-cy="Diagnosis-categorical-field"]').type('Some{downarrow}{enter}');
+    cy.get('@updateCategoricalQueryParams').should('have.been.calledWith', 'Diagnosis', {
+      id: 'https://someurl/',
+      label: 'Some Diagnosis',
+    });
+  });
+  it('Fires updateContinuousQueryParams event handler with the appropriate payload when a continuous field is selected', () => {
+    const updateContinuousQueryParams = cy.spy().as('updateContinuousQueryParams');
```
surchs: :point_up_2:

cypress/component/ResultCard.cy.tsx
```diff
@@ -0,0 +1,61 @@
...
+    cy.get('[data-cy="card-some uuid"] button')
+      .should('contain', 'ASL')
+      .should('have.class', 'bg-zinc-800');
+    cy.get('[data-cy="card-some uuid"] button')
+      .eq(1)
+      .should('contain', 'DWI')
+      .should('have.class', 'bg-red-700');
+  });
+  it('Fires onCheckboxChange event handler with the appropriate payload when the checkbox is clicked', () => {
+    const onCheckboxChange = cy.spy().as('onCheckboxChange');
```
surchs: :point_up_2:

cypress/component/ResultContainer.cy.tsx
```diff
@@ -0,0 +1,66 @@
+import ResultContainer from '../../src/components/ResultContainer';
+import { protectedResponse2 } from '../fixtures/mocked-responses';
+
+describe('ResultContainer', () => {
+  it('Displays a set of Result Cards, select all checkbox, disabled download result buttons, summary stats, and how to get data modal button', () => {
+    cy.mount(<ResultContainer result={protectedResponse2} />);
```
surchs, line 6: pretty cool that this just works, i.e. that you only need to mount the parent here and react figures out how to get the children :tada:

cypress/component/ResultContainer.cy.tsx
```diff
@@ -0,0 +1,66 @@
...
+      .should('be.visible')
+      .should('be.disabled');
+    cy.get('[data-cy="dataset-level-download-results-button"]')
+      .should('be.visible')
+      .should('be.disabled');
+    cy.get('[data-cy="how-to-get-data-modal-button"]').should('be.visible');
+  });
+  it('Selecting a dataset should enable the download result buttons', () => {
+    cy.mount(<ResultContainer result={protectedResponse2} />);
+    cy.get('[data-cy="card-http://neurobagel.org/vocab/cool-dataset-checkbox"] input').check();
```
surchs, line 23: this is minor, but as is, the test is more like a mini-integration test. To check if download button styles correctly if any dataset is selected, you could just feed the selectedDatasets prop non-empty. Here you're testing that step, but also the correct event handling for a checked dataset card, and then whether the ResultContainer processes this info correctly. So 3 steps in one. The goal would be to really just test one thing

cypress/component/ResultContainer.cy.tsx
```diff
@@ -0,0 +1,66 @@
...
+    cy.get('[data-cy="participant-level-download-results-button"]')
+      .should('be.visible')
+      .should('not.be.disabled');
+    cy.get('[data-cy="dataset-level-download-results-button"]')
+      .should('be.visible')
+      .should('not.be.disabled');
+  });
+  it('Selecting/unselecting select all datasets checkbox should check/uncheck all dataset cards', () => {
+    cy.mount(<ResultContainer result={protectedResponse2} />);
+    cy.get('[data-cy="select-all-checkbox"] input').check();
```
surchs, line 33: similar to above, this mixes the view and the logic test. Another way to describe the test is: 

if my selectedDownloads list is 
- empty, all dataset cards should be unchecked
- has some datasets, then these dataset cards should be checked

how selectedDownloads is set via the UI is not related

cypress/e2e/Form.cy.ts
```diff
@@ -0,0 +1,40 @@
+describe('App', () => {
+  it('Validates input to continuous field, displays the appropriate error, and disables the submit query button', () => {
+    cy.visit('/');
+    cy.get('[data-cy="submit-query-button"]').should('not.be.disabled');
+    cy.get('[data-cy="Minimum age-continuous-field"]').type('some text');
+    cy.get('[data-cy="Minimum age-continuous-field"] p')
+      .should('be.visible')
+      .should('contain', 'Please enter a valid number!');
```
surchs: this is cool - is there a way to assert that it takes on an "error" class?

src/components/NBDialog.tsx
```diff
@@ -12,7 +12,7 @@ function NBDialog({ open, onClose }: { open: boolean; onClose: () => void }) {
   const fullScreen = useMediaQuery(theme.breakpoints.down('md'));
 
   return (
-    <Dialog fullScreen={fullScreen} open={open} onClose={() => onClose()}>
+    <Dialog fullScreen={fullScreen} open={open} onClose={onClose} data-cy="nb-dialog">
```
surchs, line 15: this was leftover from changing the function signature for `onClose` in a previous PR, correct?

surchs (approved):
Thanks @rmanaem for this big PR, and for putting us back on trusted ground with the tests! Tried the tests locally, all runs great (and super fast now :tada:), very nice

I left some comments for clarity and some separation of concern. Take a look, and see what you think. I think we can fix some of those over time because they don't affect the tests diretly.

:cook:

.github/workflows/component-test.yaml
rmanaem: Since we made the `API_QUERY_URL` required we need it for the build to finish.

cypress/component/Navbar.cy.tsx
rmanaem: > We could just make sure that an image is loaded in this space.

Simply checking if the `img` element exists?

cypress/component/QueryForm.cy.tsx
rmanaem: Yea they're all required.

cypress/component/ResultContainer.cy.tsx
rmanaem: Indeed, no need to stub the children anymore.

src/components/NBDialog.tsx
rmanaem: No, more like a refactor after we found out that we can use this simpliefied syntax for the callback function.

.github/workflows/component-test.yaml
surchs: Got it, thanks for clarifying!

cypress/component/ResultContainer.cy.tsx
rmanaem: I see but I think the only way to manipulate the `selectedDwonloads`  is through the UI. If we want to mock the `download`, we have to change the component's logic and take the default value as a prop.

surchs: OK, let's not deal with this here / in this PR. Can think a bit more about this later

cypress/component/ResultContainer.cy.tsx
surchs: Same decision as below: let's revisit this in a later issue / PR and see if can think different thoughts about it.

cypress/component/Navbar.cy.tsx
surchs: Yeah, or maybe if it is not size zero / hasn't failed? Is there such a check?

surchs:

PR #50, opened 2024-03-05
[CI] Set up GitHub pages deployment CI

### LINKED ISSUE #41: Configure deploy on release on GH pages
We want this tool to auto-deploy on release to GH pages. To do so:

- ~[ ] Make sure this repo has our auto-release workflow setup (with labels etc)~
- [x] Configure a GH publish workflow that only triggers for a new release
- [x] (optional) have staging and preview (main but no release yet) deploys continue to run on netlify

rmanaem:
- Closes #41 

## Checklist

- [x] Tests pass
- [x] Checks pass

For new features:
- [ ] Tests have been added

For bug fixes:

### DIFF
--- a/.github/workflows/deploy.yaml
+++ b/.github/workflows/deploy.yaml
@@ -1,0 +1,49 @@
+# Adapted from sample vite workflow file: https://vitejs.dev/guide/static-deploy#github-pages
+name: deploy
+
+on:
+  release:
+    types: [published]
+  workflow_dispatch:
+
+# Sets the GITHUB_TOKEN permissions to allow deployment to GitHub Pages
+permissions:
+  contents: read
+  pages: write
+  id-token: write
+
+# Allow one concurrent deployment
+concurrency:
+  group: 'pages'
+  cancel-in-progress: true
+
+jobs:
+  deploy:
+    environment:
+      name: github-pages
+      url: ${{ steps.deployment.outputs.page_url }}
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+      - name: Set up Node
+        uses: actions/setup-node@v3
+        with:
+          node-version: 18
+          cache: 'npm'
+      - name: Create .env file
+        run: |
+          echo -e "NB_API_QUERY_URL=${{ vars.NB_API_QUERY_URL }}\nNB_IS_FEDERATION_API=${{ vars.NB_IS_FEDERATION_API }}" > .env
+      - name: Install dependencies
+        run: npm install
+      - name: Build
+        run: npm run build
+      - name: Setup Pages
+        uses: actions/configure-pages@v4
+      - name: Upload artifact
+        uses: actions/upload-pages-artifact@v3
+        with:
+          path: './dist'
+      - name: Deploy to GitHub Pages
+        id: deployment
+        uses: actions/deploy-pages@v4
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy.yaml
```diff
@@ -0,0 +1,49 @@
...
+# Sets the GITHUB_TOKEN permissions to allow deployment to GitHub Pages
+permissions:
+  contents: read
+  pages: write
+  id-token: write
+
+# Allow one concurrent deployment
+concurrency:
+  group: 'pages'
+  cancel-in-progress: true
```
surchs, line 18: :eyes: 
that's pretty cool: https://docs.github.com/en/actions/using-jobs/using-concurrency#using-concurrency-in-different-scenarios

.github/workflows/deploy.yaml
```diff
@@ -0,0 +1,49 @@
...
+
+jobs:
+  deploy:
+    environment:
+      name: github-pages
+      url: ${{ steps.deployment.outputs.page_url }}
```
surchs, line 25: I know this is from a template file, but what is going on here? This is an environment variable for the **job**, that is defined based on the output of the last **step** in this **job**? Is this environment variable used by any of the other steps in the job? And if so: how does it have any value before the last step is run?

surchs: OK, answering my own question here: **environment** is a special word from GitHub for **deployments**. And it has attributes that can be customized like 
- name
- url

So what is happening here is that we define a GitHub deployment environment based on the output of the last step in this deployment job. And what will happen is that after the deployment, this environment will show up on the GitHub UI as one of the deployment environments: https://docs.github.com/en/actions/deployment/targeting-different-environments/using-environments-for-deployment#using-an-environment

Maybe worth linking to the GitHub docs here so we remember this?

surchs: I.e. here it shows up: 
image

.github/workflows/deploy.yaml
```diff
@@ -0,0 +1,49 @@
...
+      - name: Set up Node
+        uses: actions/setup-node@v3
+        with:
+          node-version: 18
+          cache: 'npm'
+      - name: Create .env file
+        run: |
+          echo -e "NB_API_QUERY_URL=${{ vars.NB_API_QUERY_URL }}\nNB_IS_FEDERATION_API=${{ vars.NB_IS_FEDERATION_API }}" > .env
+      - name: Install dependencies
+        run: npm ci
```
surchs, line 38: :tada:

.github/workflows/deploy.yaml
```diff
@@ -0,0 +1,49 @@
...
+      name: github-pages
+      url: ${{ steps.deployment.outputs.page_url }}
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+      - name: Set up Node
+        uses: actions/setup-node@v3
+        with:
+          node-version: 18
```
surchs: should we really limit ourselves to 18 here? Pretty sure I'm running from 20 locally.

surchs (approved):
Very nice workflow @rmanaem! Thank you. I added a comment with a link from the docs - might make sense to add that as a comment to the workflow. 

:cook:

.github/workflows/deploy.yaml
rmanaem: No, you're right I think we can remove it.

surchs:

PR #51, opened 2024-03-05
[MNT] Dockerized the app

### LINKED ISSUE #48: Dockerize the app
- [x] Create a DOCKERFILE for the app
- ~[ ] Set up a CI to automatically push the docker file~
- ~[ ] Ensure it aligns with the automation in place~

rmanaem:
- Closes #48 

## Checklist

- [x] Tests pass
- [x] Checks pass

For new features:
- [ ] Tests have been added

For bug fixes:


Dockerfile
```diff
@@ -0,0 +1,15 @@
...
+
+WORKDIR /app
+
+COPY package*.json .
+
+RUN npm ci
+
+COPY . .
+
+RUN npm run build
```
surchs: could you explain how we get the ENV variables into this at runtime? i.e. when the container launches?

Dockerfile
```diff
@@ -0,0 +1,15 @@
...
+
+EXPOSE 5173
+
+ENTRYPOINT npm run preview
```
surchs: any reason to take this over the CMD? Are there any command suffixes it'd make sense to add here?

Dockerfile
```diff
@@ -0,0 +1,15 @@
...
+
+COPY package*.json .
+
+RUN npm ci
+
+COPY . .
+
+RUN npm run build
+
+EXPOSE 5173
```
surchs, line 11: why do we need this? `EXPOSE` doesn't "expose" the port: https://docs.docker.com/reference/dockerfile/#expose but is more of a doc. Is that the intent?

Dockerfile
```diff
@@ -0,0 +1,15 @@
+FROM node:latest
```
surchs: want to lock that down more with a version? I believe we use node 20 LTS. Would be good to have some consistency there

surchs (approved):
thanks @rmanaem, looks good.

Mind taking a look at the questions I posted - and then should be :cook:

Dockerfile
surchs: Just to add here: unless you figured out some special vite sauce I don't know, you'll probably have to add the bulls step as part of the entry point / cmd to ensure the runtime env variables get used here

Dockerfile
rmanaem: Similar to the Dockerfile in the old query tool using a `.env` file.

Dockerfile
rmanaem: Good call

Dockerfile
rmanaem: > any reason to take this over the CMD? Are there any command suffixes it'd make sense to add here?

I reckon we didn't want it to be overwritten.


> you'll probably have to add the bulls step as part of the entry point / cmd to ensure the runtime env variables get used here

build step? It's a couple of lines above it.

Dockerfile
rmanaem: Yea for documentation since vite uses a different port than Nuxt.

Dockerfile
surchs: Right, but this build step runs only once during image creation the way you set it up. I assume at that point you've made a bunch of static files that then get served with `npm run preview` whenever I actually launch a container from this image. So I'm wondering how you ensure that I can do this:

- docker build without providing the ENV variables (i.e. including not in my shell while building) and
- changing the ENV var every time I launch a new container from the image

In the old tool we had to build the app every time someone launched it: https://github.com/neurobagel/old-query-tool/blob/10ba7fce2b8903338386f466b81d63bcd24ee82e/Dockerfile#L15
So I'm wondering what you are doing here that makes this unnecessary

surchs: to be clear: what we did with the old app wasn't great or smart. It was just a simple way around the problem of having the ENV vars baked into the static app. There are ways around this that don't require rebuilding, but afaik they involve some form of search-replace within the built app dist folder.

rmanaem:
> ### ❌ Deploy Preview for _neurobagel-query-tool_ failed.
> Name 	Link
> 🔨 Latest commit 	[35d712e](https://github.com/neurobagel/query-tool/commit/35d712ea86a8f740c57fe1bac7d1ba815432c7a6)
> 🔍 Latest deploy log 	https://app.netlify.com/sites/neurobagel-query-tool/deploys/65e8c2851a71110008c3f119

@surchs I reckon we need to update the netlify build script.

Dockerfile
surchs: > I reckon we didn't want it to be overwritten.

That's `ENTRYPOINT`. `CMD` can be overwritten: https://docs.docker.com/reference/dockerfile/#entrypoint . But you can append commands to `ENTRYPOINT`

surchs:
Should build now: https://deploy-preview-51--neurobagel-query-tool.netlify.app/?node=All

Dockerfile
surchs: aha aha, my bad. You and I are saying the same thing. All good on ENTRYPOINT then

surchs (approved):
:tada: :cook:

surchs:

PR #80, opened 2024-03-09
[CI] Remove GH variable dependence for tests

### LINKED ISSUE #79: CI cannot run on PRs from forks because repo variables are not shared with forks
For example #63. This is annoying because it's now extra work to figure out of the tests are failing for a good reason or just because of the fact that the variables are not shared. Even worse, dependabot counts as a fork too, and based on this discussion: https://github.com/orgs/community/discussions/44322 we can expect that dependabot runs will have the same issue. 

Right now, we only use repo variables in two workflows: 
https://github.com/neurobagel/query-tool/blob/99fc5c019054aff221280b37ab1a09886721134a/.github/workflows/component-test.yaml#L17-L19

and

https://github.com/neurobagel/query-tool/blob/99fc5c019054aff221280b37ab1a09886721134a/.github/workflows/e2e-test.yaml#L22-L24

and there is only one variable that could realistically change (the API URL). I suggest we

- [ ]  get rid of the repo variables 
- [x] update the e2e workflow to hardcode the API URL
- [x] update the component test workflow to also hardcode the API URL

surchs:
Instead of referring to GH variables in the test
workflows, we just hardcode the API URL in the workflow.

This is not ideal, but better than the variable - because variables are not shared with forks.

- Closes #79 

- replace GH variable with hardcoded API URL in the workflow
- remove the GH variable (manual action, not part of the PR, will happen once merged)

## Checklist

- [x] Tests pass
- [x] Checks pass

For new features:

- [ ] Tests have been added

For bug fixes:

### DIFF
--- a/.github/workflows/component-test.yaml
+++ b/.github/workflows/component-test.yaml
@@ -16,7 +16,7 @@ jobs:
 
       - name: Create .env file
         run: |
-          echo -e "NB_API_QUERY_URL=${{ vars.NB_API_QUERY_URL }}\nNB_IS_FEDERATION_API=${{ vars.NB_IS_FEDERATION_API }}" > .env
+          echo -e "NB_API_QUERY_URL=https://federate.neurobagel.org/\nNB_IS_FEDERATION_API=true" > .env
 
       - name: build
         run: npm install && npm run build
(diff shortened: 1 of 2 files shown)


rmanaem (approved):
The changes look good. Feel free to merge after #83 🧑‍🍳 
I think the `CI` prefix would be more appropriate here wdyt?

surchs:

PR #88, opened 2024-03-13
[FIX] Fixed parsing of `NB_IS_FEDERATION_API`

### LINKED ISSUE #84: `NB_IS_FEDERATION_API` is not parsed correctly
The `NB_IS_FEDERATION_API` is currently being parsed as a `string` which in JS/TS as long as it is not an empty string is considered a truthy value therefore even when the env var is set to `false` it will be parsed as `true`.

https://github.com/neurobagel/query-tool/blob/99fc5c019054aff221280b37ab1a09886721134a/src/utils/constants.ts#L12-L13

rmanaem:
- Closes #84 

## Checklist

- [x] Tests pass
- [x] Checks pass

For new features:

- [ ] Tests have been added

For bug fixes:

### DIFF
--- a/src/components/QueryForm.tsx
+++ b/src/components/QueryForm.tsx
@@ -95,14 +95,14 @@ function QueryForm({
           />
         </div>
       )}
-      <div className={isFederationAPI && 'row-start-2'}>
+      <div className={isFederationAPI ? 'row-start-2' : ''}>
         <ContinuousField
           helperText={minAgeExceedsMaxAge ? '' : minAgeHelperText}
           label="Minimum age"
           onFieldChange={updateContinuousQueryParams}
         />
       </div>
-      <div className={isFederationAPI && 'row-start-2'}>
+      <div className={isFederationAPI ? 'row-start-2' : ''}>
         <ContinuousField
           helperText={minAgeExceedsMaxAge ? '' : maxAgeHelperText}
           label="Maximum age"
--- a/src/utils/constants.ts
+++ b/src/utils/constants.ts
@@ -9,8 +9,7 @@ export const nodesURL: string = baseAPIURL.endsWith('/')
   ? `${baseAPIURL}nodes/`
   : `${baseAPIURL}/nodes/`;
 
-export const isFederationAPI: string =
-  import.meta.env.NB_IS_FEDERATION_API === undefined ? true : import.meta.env.NB_IS_FEDERATION_API;
+export const isFederationAPI: boolean = import.meta.env.NB_IS_FEDERATION_API === 'true';
 
 export const sexes: { [key: string]: string } = {
   male: 'snomed:248153007',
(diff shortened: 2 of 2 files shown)


src/components/QueryForm.tsx
```diff
@@ -95,14 +95,14 @@ function QueryForm({
           />
         </div>
       )}
-      <div className={isFederationAPI && 'row-start-2'}>
+      <div className={isFederationAPI ? 'row-start-2' : ''}>
```
surchs, line 98: make computed variable with reasonable name?

surchs: So we don't have to duplicate this logic here

src/utils/constants.ts
```diff
@@ -9,8 +9,7 @@ export const nodesURL: string = baseAPIURL.endsWith('/')
   ? `${baseAPIURL}nodes/`
   : `${baseAPIURL}/nodes/`;
 
-export const isFederationAPI: string =
-  import.meta.env.NB_IS_FEDERATION_API === undefined ? true : import.meta.env.NB_IS_FEDERATION_API;
+export const isFederationAPI: boolean = import.meta.env.NB_IS_FEDERATION_API === 'true';
```
surchs: I believe this is changing the default behaviour from "undefined -> true" to "undefined -> false". Can you please

1. Confirm this is not the case / otherwise fix
2. Add explicit tests for 
  - env.NB_IS_FEDERATION_API = undefined
  - env.NB_IS_FEDERATION_API = true
  - env.NB_IS_FEDERATION_API = false

so we know for sure and don't have to worry

surchs (commented):
thanks @rmanaem but this seems to change default behaviour. Could you please confirm and add tests

src/components/QueryForm.tsx
rmanaem: Moving this logic outside of the layout would result in reduced readability.

surchs: could you explain how that is? If the variable was called `form_start_row`, how would that be less readable?

surchs: I am now convinced of the opposite of what I was asking for

src/utils/constants.ts
rmanaem: I've updated the fix so the default remains unchanged.
I wasn't able to find a similar workaround to what we used in Vue for overwriting the env vars. Here is the section of the Cypress docs that states `mount` is frameworks-specific hence why we may not be able to duplicate what we had in Vue: https://docs.cypress.io/api/commands/mount#Creating-a-New-cymount-Command

surchs (approved):
:cook:

surchs:
before you merge @rmanaem, mind updating the commit message in the squashed commit:

> [Modified the fix so the default is false](https://github.com/neurobagel/query-tool/pull/88/commits/be94f6a399601fa30218a18d39f4e9bf58bfbd5c)

to instead say `true` - I blieve that's what you meant?

rmanaem:
> to instead say `true` - I blieve that's what you meant?

Yes, good catch.

surchs:

PR #97, opened 2024-03-26
[REF] Enable handling of partial node response success

### LINKED ISSUE #60: Handle partial node query successes when all nodes are selected
See https://github.com/neurobagel/federation-api/blob/350469a3a72d9ac6278d8371116c195b95a1b276/app/api/models.py#L57-L62 for what the new response from the federation API should look like.

Note that node API responses have not changed their shape at all.

### LINKED ISSUE #96: Create an e2e test to ensure the app loads correctly
Right now, all the e2e tests in https://github.com/neurobagel/query-tool/blob/c890a4a5a2c221993b8240ff106b44847df4ba2c/cypress/e2e/APIRequests.cy.ts are concerned with correctly handling certain failure modes. However, the one test that is supposed to check the happy path is very broad - it only checks that the query goes to the correct URL.

Instead we should have a dedicated test to assert a scenario that
- does not show any errors
- correctly populates the Diagnosis and assessment options
- shows a list of nodes
- AND all of the above without relying on a specific running API (i.e. all should be mocked)

Right now, I can completely change the API response shema in #60 and not have a single e2e test fail. That's no good!

surchs:
- Closes #60 

- Refactored mocked API responses into
  - query response section
  - attribute response section
- reused most of the responses to make the mocks less cluttered / easier to understand
- added an e2e test for the "happy path" of the app to assert that no warnings are shown
  - closes #96
- expanded the API request tests to cover the new partial and complete success and failure cases for attribute requests
- updated the types in the ResultsContainer

## Checklist

- [x] Tests pass
- [x] Checks pass

For new features:

- [ ] Tests have been added

For bug fixes:


src/App.tsx
```diff
@@ -313,11 +327,7 @@ function App() {
           />
         </div>
         <div className="col-span-3">
-          <ResultContainer
-            result={
-              result ? result.sort((a, b) => a.dataset_name.localeCompare(b.dataset_name)) : null
```
rmanaem, line 318: Curious, why remove the sorting?

cypress/e2e/Form.cy.ts
```diff
@@ -17,10 +19,13 @@ describe('App', () => {
...
-    cy.intercept({
-      method: 'GET',
-      url: '/attributes/nb:Diagnosis',
-    }).as('getDiagnosisOptions');
+    cy.intercept(
+      {
+        method: 'GET',
+        url: '/attributes/nb:Diagnosis',
+      },
+      diagnosisOptions
```
rmanaem, line 47: Good call!

cypress/e2e/APIRequests.cy.ts
```diff
@@ -46,38 +61,103 @@ describe('API request', () => {
...
+        method: 'GET',
+        url: '/attributes/nb:Assessment',
+      },
+      failedAssessmentToolOptions
+    ).as('getAssessmentToolOptions');
+    cy.visit('/');
+    cy.wait(['@getNodes', '@getDiagnosisOptions', '@getAssessmentToolOptions']);
+  });
+  it('Shows error toast for failed Assessment tool options', () => {});
+  it('Shows error toast for failed Diagnosis options', () => {});
```
rmanaem: May have forgotten to address the `failure` case.

cypress/e2e/APIRequests.cy.ts
```diff
@@ -46,38 +61,103 @@ describe('API request', () => {
...
+      },
+      mixedResponse
+    ).as('call');
+    cy.intercept(
+      {
+        method: 'GET',
+        url: '/nodes/',
+      },
+      nodeOptions
+    ).as('getNodes');
```
rmanaem, line 169: Could also mock the attribute related requests here as well.

rmanaem (changes requested):
Thanks for taking over this @surchs!
The refactor looks clean and nicely done.
🍒  One thing I noticed is missing is handling `failure` case for attributes and query response.

rmanaem:
> added an e2e test for the "happy path" of the app to assert that no warnings are shown 

Is this referring to this test?
https://github.com/neurobagel/query-tool/blob/3994a0b2859be3f5abca4ee6493793216aed8944/cypress/e2e/APIRequests.cy.ts#L14-L42

src/App.tsx
surchs: Ah, yeah, should probably bring that back! Good call. I removed it because the type changed so sorting gets more complex (has to happen on a specific key inside the array) - but still feasible.

cypress/e2e/APIRequests.cy.ts
surchs: Just to be clear: you mean I forgot to add a test for this case, correct? Not that the handling of such a case is missing from the app itself?

surchs:
> > added an e2e test for the "happy path" of the app to assert that no warnings are shown
> 
> Is this referring to this test?
> 
> https://github.com/neurobagel/query-tool/blob/3994a0b2859be3f5abca4ee6493793216aed8944/cypress/e2e/APIRequests.cy.ts#L14-L42

Yes, exactly!

cypress/e2e/APIRequests.cy.ts
rmanaem: I believe it's both.

rmanaem: From the issue description

> complete failure
   none of the nodes have responded
    -> do not show any results, instead show an error message in the results area
    -> also show an error toast and say that all nodes have failed.

surchs: OK, thanks for the hint @rmanaem. Turns out this was a bigger change - so probably best to review again!

cypress/e2e/APIRequests.cy.ts
```diff
@@ -1,83 +1,318 @@
...
+      {
+        method: 'GET',
+        url: '/attributes/nb:Assessment',
+      },
+      assessmentToolOptions
+    ).as('getAssessmentToolOptions');
+    cy.visit('/');
+    cy.wait(['@getNodes', '@getDiagnosisOptions', '@getAssessmentToolOptions']);
+  });
+  it('Intercepts the request sent to the APpI and asserts over the request url', () => {
```
rmanaem:
```suggestion
  it('Intercepts the request sent to the API and asserts over the request url', () => {
```

cypress/e2e/APIRequests.cy.ts
```diff
@@ -1,83 +1,318 @@
...
+      },
+      diagnosisOptions
+    ).as('getDiagnosisOptions');
+    cy.intercept(
+      {
+        method: 'GET',
+        url: '/attributes/nb:Assessment',
+      },
+      assessmentToolOptions
     ).as('getAssessmentToolOptions');
```
rmanaem: Duplicate code

cypress/e2e/APIRequests.cy.ts
```diff
@@ -1,83 +1,318 @@
...
     cy.visit('/');
-    cy.wait('@getDiagnosisOptions');
-    cy.get('.notistack-SnackbarContainer').should('contain', 'No Diagnosis options were available');
-    cy.get('.notistack-SnackbarContainer').should(
-      'contain',
-      'No Assessment tool options were available'
-    );
+    cy.wait(['@getNodes', '@getDiagnosisOptions', '@getAssessmentToolOptions']);
+  });
+  it('Shows a warning for nodes that did not return any results', () => {
```
rmanaem:
```suggestion
  it('Shows a warning for nodes that failed to return any results', () => {
```

cypress/e2e/APIRequests.cy.ts
```diff
@@ -1,83 +1,318 @@
...
+      diagnosisOptions
     ).as('getDiagnosisOptions');
     cy.intercept(
       {
         method: 'GET',
         url: '/attributes/nb:Assessment',
       },
-      { statusCode: 500 }
+      assessmentToolOptions
     ).as('getAssessmentToolOptions');
```
rmanaem: Duplicate code

src/components/ResultContainer.tsx
```diff
@@ -156,14 +157,28 @@ function ResultContainer({ result }: { result: Result[] | null }) {
...
           Click &apos;Submit Query&apos; for results
         </Typography>
       );
     }
-    if (result.length === 0) {
+
+    if (response.nodes_response_status === 'failure') {
+      return (
+        <div data-cy="failed-result-container-view">
+          <Typography variant="h5" className="text-red-500">
```
rmanaem: This is more of a personal preference, I think using the default black color is better here as opposed to red especially since we're already displaying a red toast in case of failure.

rmanaem (changes requested):
:cherries:

src/components/ResultContainer.tsx
surchs: Agree, was considering this as well because red looks scary!

rmanaem (approved):
:cook:

surchs:

PR #205, opened 2024-07-14
[ENH] Implemented OIDC flow that results in a query sent to the f-API

### LINKED ISSUE #197: Implement OIDC flow that results in a query sent to the f-API including access and ID tokens
(this assumes the **implicit** OAuth2 scheme)
- [x] Query tool will redirect user to authenticate using IdP
- [x] Query tool should get back a code from IdP (via a redirect URL)
- [x] Query tool provides code back to IdP to get an **access token** and ID token in exchange
~- [ ] Query tool needs to verify received ID token's validity using IdP's public keys (existing JWT libraries have functionality to do this)~
- [ ] (Maybe) Query tool can then exchange **access token** for user profile pic from identity resource API
  - or, alternatively, query tool just displays the IdP user ID directly
- [x] Query tool now displays the user profile pic from IdP, and can also display the name (taken from the ID token probably)
- [x] When user submits a query, query tool sends a request to f-API, along with the ID & access tokens

## Questions
- Do we want to make it so the user doesn't have to login on every refresh? i.e., do we want to use a session?
- Why doesn't Google Authentication work as expected with `127.0.0.1`?
  - Relevant docs: https://developers.google.com/identity/protocols/oauth2/javascript-implicit-flow#origin-validation

rmanaem:
- Closes #197 

-
-

## Checklist

- [x] Tests pass
- [x] Checks pass

For new features:
- [x] Tests have been added

For bug fixes:


cypress/component/AuthDialog.cy.tsx
```diff
@@ -0,0 +1,26 @@
...
+};
+
+describe('ContinuousField', () => {
+  it('Displays a MUI dialog with the title and "sing in with google" button', () => {
+    cy.mount(
+      <GoogleOAuthProvider clientId="mock-client-id">
+        {' '}
+        <AuthDialog isLoggedIn={props.isLoggedIn} onAuth={props.onAuth} />
+      </GoogleOAuthProvider>
+    );
```
surchs, line 16: Is this a special trick to get the auth provider to run inside the e2e test? Shouldn't hold up the PR, but we should note that down in our notes

cypress/component/AuthDialog.cy.tsx
```diff
@@ -0,0 +1,26 @@
...
+    cy.get('[data-cy="auth-dialog"]').should('be.visible');
+    cy.get('[data-cy="auth-dialog"]').should(
+      'contain',
+      'Please login using one of the following before proceeding'
+    );
+    cy.get('[data-cy="auth-dialog"]').within(() => {
+      cy.contains('Sign in with Google');
```
surchs: Pretty interesting error I ran into here: The "Sign in with Google" uses the language of the country from where you make the request. So the test would fail in Germany because here the string will be 
image

Something to keep in mind for e2e tests. For now I'll switch the check to just "Google"

src/App.tsx
```diff
@@ -308,19 +338,40 @@ function App() {
...
     setLoading(false);
   }
 
+  function login(credential: string | undefined) {
+    setIsLoggedIn(true);
+    const jwt: GoogleJWT = credential ? jwtDecode(credential) : ({} as GoogleJWT);
+    setIDToken(credential);
+    setName(jwt.given_name);
+    setProfilePic(jwt.picture);
+  }
```
surchs, line 352: That'd be a good candidate for a custom hook in the future afaict - so we can subscribe to the global "logged in" status with all components that need it.

src/components/Navbar.tsx
```diff
@@ -20,6 +40,13 @@ function Navbar() {
       });
   }, []);
 
+  const handleClick = (event: React.MouseEvent<HTMLElement>) => {
+    setAnchorEl(event.currentTarget);
+  };
+  const handleClose = () => {
+    setAnchorEl(null);
+  };
```
surchs, line 48: We should revisit this bit soon and see if by refactoring the whole login section into its own component we can get rid of the custom click handling. Could you open an issue please.

src/main.tsx
```diff
@@ -12,11 +14,15 @@ const router = createBrowserRouter([
...
     {/* CSS injection order for MUI and tailwind: https://mui.com/material-ui/guides/interoperability/#tailwind-css */}
     <StyledEngineProvider injectFirst>
       <RouterProvider router={router} />
     </StyledEngineProvider>
   </React.StrictMode>
 );
+
+ReactDOM.createRoot(document.getElementById('root')!).render(
+  enableAuth ? <GoogleOAuthProvider clientId={clientID}>{app}</GoogleOAuthProvider> : app
+);
```
surchs, line 28: neat!

src/App.tsx
```diff
@@ -42,6 +47,11 @@ function App() {
   const [imagingModality, setImagingModality] = useState<FieldInput>(null);
   const [searchParams, setSearchParams] = useSearchParams();
 
+  const [isLoggedIn, setIsLoggedIn] = useState<boolean>(false);
+  const [name, setName] = useState<string>('');
+  const [profilePic, setProfilePic] = useState<string>('');
+  const [IDToken, setIDToken] = useState<string | undefined>('');
+
```
surchs, line 54: could be good candidate for single state object and reducer

src/App.tsx
```diff
@@ -53,6 +63,16 @@ function App() {
...
+  const action = (snackbarId: SnackbarKey) => (
+    <IconButton
+      onClick={() => {
+        closeSnackbar(snackbarId);
+      }}
+    >
+      <CloseIcon className="text-white" />
+    </IconButton>
+  );
+
```
surchs, line 75: nice - we probably want to also refactor this soon into it's own component though

src/App.tsx
```diff
@@ -308,19 +338,40 @@ function App() {
...
+  }
+
+  function logout() {
+    googleLogout();
+    setIsLoggedIn(false);
+    setIDToken('');
+    setName('');
+    setProfilePic('');
+  }
+
```
surchs, line 361: I think that's also a good reducer candidate for a cleanup / refactor

src/App.tsx
```diff
@@ -308,19 +338,40 @@ function App() {
...
     }
     setLoading(false);
   }
 
+  function login(credential: string | undefined) {
+    setIsLoggedIn(true);
+    const jwt: GoogleJWT = credential ? jwtDecode(credential) : ({} as GoogleJWT);
+    setIDToken(credential);
+    setName(jwt.given_name);
+    setProfilePic(jwt.picture);
```
surchs, line 351: really cool!

src/App.tsx
```diff
@@ -280,19 +302,27 @@ function App() {
...
     const url: string = constructQueryURL();
     try {
-      const response = await axios.get(url);
+      const response = await axios.get(url, {
+        headers: {
+          Authorization: `Bearer ${IDToken}`,
+          'Content-Type': 'application/json',
+        },
+      });
       // TODO: remove this branch once there is no more non-federation option
```
surchs, line 311: If auth is NOT enabled, does this just send the header anyway, but with an empty ID token? That's probably alright for now, but we should revisit!

surchs (approved):
Very cool PR @rmanaem! :tada: Very exciting new functionality!

I read through it and it looks and feels really nice. As you said, there are some things we can refactor / clean up soon, but the logic is really clear and the functionality works cleanly for me. I left some comments here so we can go back and look at them together. But for now I will merge this!

:cook:

PR #214, opened 2024-07-16
[ENH] Made authentication optional on the UI

### LINKED ISSUE #212: Make authentication optional on UI
Once https://github.com/neurobagel/api/issues/322 is resolved, node API instances can also enable authentication.
We'd like the query tool to be able to work with both federation API and node API regardless of whether they have authentication enabled or not.
Although the authentication flow can be toggled on deployment, we want to make the authentication optional on the UI front to allow users to make queries to APIs that do not require authentication freely. 
The goal here is to allow the user to send the query without having to authenticate, how APIs handle the presence or absence of the ID token when receiving and sending out requests and how they respond to requests is their responsibility.

- [x] User should be able to close the modal without authenticating
- [x] User should be able to access the modal again once it's closed 
   - [x] Preferably by selecting the profile menu
- [x] Requests should go out to the API whether user has been authenticated or not

rmanaem:
- Closes #212 

-
-

## Checklist

- [x] Tests pass
- [x] Checks pass

For new features:
- [x] Tests have been added

For bug fixes:

### DIFF
--- a/cypress/component/AuthDialog.cy.tsx
+++ b/cypress/component/AuthDialog.cy.tsx
@@ -2,16 +2,15 @@ import { GoogleOAuthProvider } from '@react-oauth/google';
 import AuthDialog from '../../src/components/AuthDialog';
 
 const props = {
-  isLoggedIn: false,
   onAuth: () => {},
+  onclose: () => {},
 };
 
 describe('ContinuousField', () => {
   it('Displays a MUI dialog with the title and "sing in with google" button', () => {
     cy.mount(
       <GoogleOAuthProvider clientId="mock-client-id">
-        {' '}
-        <AuthDialog isLoggedIn={props.isLoggedIn} onAuth={props.onAuth} />
+        <AuthDialog open onClose={props.onclose} onAuth={props.onAuth} />
       </GoogleOAuthProvider>
     );
     cy.get('[data-cy="auth-dialog"]').should('be.visible');
@@ -19,5 +18,6 @@ describe('ContinuousField', () => {
     cy.get('[data-cy="auth-dialog"]').within(() => {
       cy.contains('Google');
     });
+    cy.get('[data-cy="auth-dialog-close-button"]').should('be.visible');
   });
 });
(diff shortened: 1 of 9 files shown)


cypress/component/AuthDialog.cy.tsx
```diff
@@ -2,22 +2,22 @@ import { GoogleOAuthProvider } from '@react-oauth/google';
 import AuthDialog from '../../src/components/AuthDialog';
 
 const props = {
-  isLoggedIn: false,
   onAuth: () => {},
+  onclose: () => {},
```
surchs:
```suggestion
  onClose: () => {},
```
consistent capitalizartion?

cypress/component/Navbar.cy.tsx
```diff
@@ -1,23 +1,34 @@
...
+      />
+    );
     cy.get("[data-cy='navbar']").should('be.visible');
     cy.get("[data-cy='navbar'] img").should('exist');
     cy.get("[data-cy='navbar'] h5").should('contain', 'Neurobagel Query');
     cy.get("[data-cy='navbar'] p").should(
       'contain',
       'Define and find cohorts at the subject level'
     );
+    cy.get("[data-cy='navbar'] a").eq(0).find('svg').should('be.visible');
```
surchs, line 30: could you add a comment here on what this is? I'm assuming this checks for the user icon? Maybe we can give the user icon its own data-cy name?

src/App.tsx
```diff
@@ -345,6 +346,7 @@ function App() {
 
   function login(credential: string | undefined) {
     setIsLoggedIn(true);
+    setOpenAuthDialog(false);
```
surchs, line 349: So now we default to not asking for login, even if auth is enabled?
I think that's fine, but may lead to the odd situation where you send a query and get no replies / errors because you forgot to log in. We should probably address that case soon as well.

surchs (approved):
Looks good @rmanaem, thanks for the PR!

left some comments, but :cook:

cypress/component/Navbar.cy.tsx
rmanaem: This checks for the documentation icon.

src/App.tsx
rmanaem: I believe we should address that together with the APIs. If some API has Auth enabled and gets no ID token in the request, it should return an error message that we can then display to the user using toasts on the query tool.

PR #245, opened 2024-08-06
[ENH] Removed `NB_IS_FEDERATION_API` environment variable

### LINKED ISSUE #239: Remove `NB_IS_FEDERATION_API` environment variable
We've decided to no longer make the API type that the query tool points to configurable, so this variable is no longer needed.

rmanaem:
- Closes #239 

- Removed code related to `NB_IS_FEDERATION_API`
- Updated workflow config files
- Updated README.md

## Checklist

- [x] Tests pass
- [x] Checks pass

For new features:
- [ ] Tests have been added

For bug fixes:

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,21 +47,16 @@ but before proceeding with either you need to set the environment variables.
 
 ### Mandatory configuration
 
-| Environment variable   | Type    | Required                                 | Default value if not set | Example                                                   |
-| ---------------------- | ------- | ---------------------------------------- | ------------------------ | --------------------------------------------------------- |
-| `NB_API_QUERY_URL`     | string  | Yes                                      | -                        | `https://federate.neurobagel.org/`                        |
-| `NB_IS_FEDERATION_API` | boolean | No                                       | `true`                   | `true`                                                    |
-| `NB_QUERY_URL_PATH`    | string  | No                                       | `/`                      | `/query/`                                                 |
-| `NB_ENABLE_AUTH`       | boolean | No                                       | `false`                  | `false`                                                   |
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -82,16 +77,15 @@ To set environment variables, create a `.env` file in the root directory and add
 
 ```bash
 NB_API_QUERY_URL=http://localhost:8000/
-NB_IS_FEDERATION_API=false # For node API
 ```
 
-if you're using the remote (in this example federation) api, your `.env` file would look something like this:
```
surchs, line 88: good find

README.md
```diff
@@ -82,16 +77,15 @@ To set environment variables, create a `.env` file in the root directory and add
...
+if you're using the remote api, your `.env` file would look something like this:
 
 ```bash
 NB_API_QUERY_URL=https://federate.neurobagel.org/
 ```
 
-if you're using a federation api with authentication, your `.env` file would look something like this:
+if you're using a remote api with authentication, your `.env` file would look something like this:
```
surchs, line 88: in this case, why not keep `federation`?  we now know it has to always be a f-API.

surchs (approved):
Thanks for the PR @rmanaem, really nice to see how much simpler everything is getting with this change!

Very :cherries: but I find `[ENH]` a bit odd for this. Technically we're taking a feature away. How about `[MNT]` instead? Or maybe we need a new prefix for feature deprecation / removal?

I left a comment, and then :cook:

README.md
rmanaem: Since we only support `federation`, the condition here (if) no longer makes sense.

PR #322, opened 2024-10-22
[ENH] Implemented logic to allow for querying pipeline name and version

### LINKED ISSUE #282: Allow querying for pipeline name and version
- [x] Add two new dropdown fields for pipeline name and version (should have an extra/default option meaning "all available versions")
  - [x] Once user selects a pipeline name, the version dropdown options should update to be available vers for that pipeline
- [x] #293
- [x] Add pipeline name and version to downloadable results files

### LINKED ISSUE #293: Display available pipeline names and versions to matching dataset card pipelines available for the dataset

rmanaem:
- Closes #282 
- Closes #293 

## Checklist

- [x] Tests pass
- [x] Checks pass

For new features:
- [x] Tests have been added

For bug fixes:


src/utils/constants.ts
```diff
@@ -1,13 +1,10 @@
...
-  : `${baseAPIURL}/query?`;
-export const attributesURL: string = baseAPIURL.endsWith('/')
-  ? `${baseAPIURL}attributes/`
-  : `${baseAPIURL}/attributes/`;
-export const nodesURL: string = baseAPIURL.endsWith('/')
-  ? `${baseAPIURL}nodes`
-  : `${baseAPIURL}/nodes`;
+export const baseAPIURL: string = import.meta.env.NB_API_QUERY_URL.endsWith('/')
+  ? import.meta.env.NB_API_QUERY_URL
+  : `${import.meta.env.NB_API_QUERY_URL}/`;
```
surchs: I think it's still cleaner here to read the ENV variable just once, into a local variable,
and then do the slash handling in the way you did before. 
```suggestion
const userAPIURL: string = import.meta.env.NB_API_QUERY_URL
export const baseAPIURL: string = userAPIURL.endsWith('/')
  ? userAPIURL
  : `${userAPIURL}/`;
```
or something like that

src/utils/constants.ts
```diff
@@ -1,13 +1,10 @@
...
+
+export const queryURL: string = `${baseAPIURL}query?`;
```
surchs, line 7: I'm sure we looked at that, but making sure URL variables have slashes and so ...
Isn't there a tool for that? People use tools to left-pad strings ...

src/App.tsx
```diff
@@ -76,35 +81,39 @@ function App() {
   );
 
   useEffect(() => {
-    async function getAttributes(dataElementURI: string) {
+    async function getAttributes(NBClass: string, dataElementURI: string) {
```
surchs: How about `NBResource` instead of `Class`. Class is a bit generic for me

```suggestion
    async function getAttributes(NBResource: string, dataElementURI: string) {
```

src/App.tsx
```diff
@@ -148,6 +165,58 @@ function App() {
...
+      } catch {
+        return [];
+      }
+    }
+    // Get pipeline versions if
+    // 1. A pipeline has been selected (this implementation only works for single value for pipeline name field)
+    // 2. This is the first time its being selected (i.e., we haven't retrieved pipeline versions before)
+    if (
+      pipelineName !== null &&
+      !Array.isArray(pipelineName) &&
```
surchs, line 208: Is this line checking whether I have selected multiple pipelines? Can I even do that right now?

src/components/ResultContainer.tsx
```diff
@@ -115,6 +117,13 @@ function ResultContainer({ response }: { response: QueryResponse | null }) {
                   subject.num_matching_phenotypic_sessions,
                   subject.num_matching_imaging_sessions,
                   subject.image_modal?.join(', '),
+                  subject.completed_pipelines
+                    ? Object.entries(subject.completed_pipelines)
+                        .flatMap(([name, versions]) =>
```
surchs: oh cool, didn't know flatMap

src/components/ResultContainer.tsx
```diff
@@ -115,6 +117,13 @@ function ResultContainer({ response }: { response: QueryResponse | null }) {
...
+                          (versions as string[]).map((version: string) => `${name} ${version}`)
+                        )
+                        .join(', ')
+                    : {},
                 ].join('\t')
```
surchs: I think this is a bit much complexity for an inline piece of code here. 
I get that you're constructing the pipeline+version string array here
but compared to all the other lines above it is much harder to read / understand.

Factor out into a function?

src/components/ResultContainer.tsx
```diff
@@ -138,6 +148,11 @@ function ResultContainer({ response }: { response: QueryResponse | null }) {
               res.dataset_portal_uri,
               res.num_matching_subjects,
               res.image_modals?.join(', '),
+              res.available_pipelines
+                ? Object.entries(res.available_pipelines).flatMap(([name, versions]) =>
+                    versions.map((version) => `${name} ${version}`).join(', ')
+                  )
+                : {},
             ].join('\t')
```
surchs: seems this section could make use of a shared, generic function that also addresses the unwrapping above

src/components/ResultCard.tsx
```diff
@@ -37,20 +43,46 @@ const ResultCard = memo(
...
+              title={
+                <Typography variant="body1">
+                  {Object.entries(pipelines)
+                    .flatMap(([name, versions]) =>
+                      versions.map((version) => `${name.slice(65)} ${version}`)
+                    )
+                    .map((pipeline) => (
+                      <Divider>{pipeline}</Divider>
+                    ))}
+                </Typography>
```
surchs, line 65: another section for unwrapping pipeline names into text that (on first glance)
shares quite a lot of overlap with code in `ResultContainer.tsx` and might be 
able to share a generic function, also to improve readability here!

src/components/QueryForm.tsx
```diff
@@ -180,7 +194,50 @@ function QueryForm({
           inputValue={imagingModality}
         />
       </div>
-      <div className="row-start-10">
+      <div className="col-span-2 row-start-9">
+        <CategoricalField
+          label="Pipeline name"
+          options={Object.keys(pipelines).map((p) => ({
+            label: p.slice(3),
```
surchs: what am I slicing here? 
Ah, I guess `nb:`, yes? 
That's real tricky to remember and needs a comment. 
Additionally we could call `p` -> `pipelineURI` to 
assist with that memory

src/components/QueryForm.tsx
```diff
@@ -180,7 +194,50 @@ function QueryForm({
...
+              disabled
+            />
+          </div>
+        </Tooltip>
+      ) : (
+        <div className="col-span-2 row-start-10">
+          <CategoricalField
+            label="Pipeline version"
+            options={[
+              { label: 'All', id: '' },
```
surchs: minor: I don't think an "All" option is needed here. The only other place we use that is for the node-dropdowns.
Everywhere else "empty / unselected" means All. And I think that's fine.

Unrelated: does `id` not need to be set to something other than emptystring?

src/components/CategoricalField.tsx
```diff
@@ -12,7 +12,10 @@ function CategoricalField({
   return (
     <Autocomplete
       data-cy={`${label}-categorical-field`}
-      options={options.sort((a, b) => a.label.localeCompare(b.label))}
+      options={options.sort((a, b) => {
+        if (a.label === 'All') return -1;
```
surchs: Here too: "All" introduces a bit of extra complexity without being very necessary. 
I think we could get rid of "All" and just use `Null`/unselected internally to mean the same.

Unless of course you added "All" because what I suggest is much harder. Then I'd add a comment to explain that.

cypress/component/ResultCard.cy.tsx
```diff
@@ -24,6 +30,7 @@ describe('ResultCard', () => {
         datasetTotalSubjects={props.datasetTotalSubjects}
         numMatchingSubjects={props.numMatchingSubjects}
         imageModals={props.imageModals}
+        pipelines={props.pipelines}
```
surchs, line 33: please also update test description to include "pipeline"
Generally: this test is a bit "Everything", we may want to start testing things individually and use `beforeEach` to keep the setup the same. Otherwise the test descriptions become half of an essay 
and the tests meaningless

surchs (approved):
Thanks for this big PR @rmanaem! Really cool to have the functionality!
Really impressive work with the test suites as well - I wonder if we have to start thinking about making the test-update burden lower. 

I left a couple of comments, mainly for clarity. Please take a look at those and decide what can be now vs what should be in an issue for after the PR. 

With that: :cook:

src/App.tsx
rmanaem: Kinda, it's way to get around the `FieldInputOption []` which is part of the `FieldInput` to prevent the type error.

src/components/CategoricalField.tsx
rmanaem: I believe we discussed and decided to explicitly add `All` for pipeline versions during stand up.

rmanaem: To clarify this bit of code was added as a result of including `All` in the pipeline versions and the need to have condition to order `All` at the top of the dropdown options.

src/components/QueryForm.tsx
rmanaem: > minor: I don't think an "All" option is needed here. The only other place we use that is for the node-dropdowns.
Everywhere else "empty / unselected" means All. And I think that's fine.

I believe we discussed and decided to explicitly add All for pipeline versions during stand up.

> Unrelated: does id not need to be set to something other than emptystring?

No, we filter out empty strings when generating the query url.

src/utils/constants.ts
rmanaem: I don't think we did. Will check.

rmanaem: It seems like this is the least verbose approach since other tools/libraries like `url-join`, `ulijs` are designed for more complex scenarios and are unnecessary for this purpose.

src/components/ResultCard.tsx
rmanaem: I don't think this one could use the same generic function used in `ResultContainer.tsx` as it involves different logic.

PR #357, opened 2024-11-17
[ENH] Modified the result tables for humans and machines

### LINKED ISSUE #136: Change TSV column names `Num{Phenotypic,Imaging}Sessions` to `NumMatching{Phenotypic,Imaging}Sessions`
The counts reflect the number of session matching the query, not the total per subject, so we should update the column names in the results TSVs to avoid confusion (the API results also include "matching" in the corresponding field names)

### LINKED ISSUE #356: make the participants-results table nice for humans and machines
The query tool let's you download information about cohorts you find. We have two kinds of users in mind for this information:

1. human users, who likely want to look at the downloaded information to learn more about their cohort
2. machines or tools who want to use the information to provide services to their own users, e.g. to download a cohort to the users computer (or create symlinks https://github.com/neurobagel/planning/issues/188)

Right now our downloadable files do a little bit of both, and neither very nicely (e.g. include many fully expanded IRIs etc). We mainly want to make sure that 

- [x] there is one file that includes information about the cohort in a nice-for-humans format
- [x] there is *another* file that includes enough information for tools like cohort-creator to keep doing what they are doing

And to achieve this, we are willing to have a bit of duplication in these files.

Here is a draft of both formats for us to discuss: https://docs.google.com/spreadsheets/d/1rLgbzZv1AqgYTGglI3yuQVdOM1bbYvvY8ijmbTXsfHg/edit?usp=sharing

TODO

- [x] Discuss and agree on a first format for the human and machine readable files
- [x] Pick an intuitive name for both files
- [x] Implement the new file formats in the query tool
- [x] Update the query tool examples in neurobagel/examples (discuss / decide what to do with the old examples)
- [ ] Ping Remi in cohort-creator land about the upcoming change to the format of these files
- [ ] Deploy / Release the new query tool version

rmanaem:
- Closes #356 
- Closes #136

## Checklist

- [x] Tests pass
- [x] Checks pass
- [x] If the PR changes the participant-level and/or the dataset-level result file, the [`query-tool-results` files](https://github.com/neurobagel/neurobagel_examples/tree/main/query-tool-results) in the  [neurobagel_examples repo](https://github.com/neurobagel/neurobagel_examples/) have been regenerated

For new features:
- [x] Tests have been added

For bug fixes:

## Summary by Sourcery

Modify the result tables to improve readability for both humans and machines by updating the ResultContainer component and adjusting the TSV download format. Enhance the Cypress tests to align with these changes.

Enhancements:
- Improve the ResultContainer component to handle additional options for diagnosis and assessment, and enhance the conversion of URIs to labels for better readability.

Tests:
- Update Cypress tests to reflect changes in the result download process, including new button identifiers and expected file content.

### DIFF
--- a/src/components/DownloadResultButton.tsx
+++ b/src/components/DownloadResultButton.tsx
@@ -18,7 +18,7 @@ function DownloadResultButton({
       onClick={() => handleClick(identifier)}
       disabled={disabled}
     >
-      Download {identifier} result
+      Download {identifier.split('-').join(' ')} result
     </Button>
   );
 
--- a/src/components/DownloadResultButton.tsx
+++ b/src/components/DownloadResultButton.tsx
@@ -18,7 +18,7 @@ function DownloadResultButton({
       onClick={() => handleClick(identifier)}
       disabled={disabled}
     >
-      Download {identifier} Result
+      Download {identifier} results
     </Button>
   );
 
--- a/src/components/DownloadResultButton.tsx
+++ b/src/components/DownloadResultButton.tsx
@@ -18,7 +18,7 @@ function DownloadResultButton({
       onClick={() => handleClick(identifier)}
       disabled={disabled}
     >
-      Download {identifier} results
+      Download {identifier} result
     </Button>
   );
 
--- a/src/components/GetDataDialog.tsx
+++ b/src/components/GetDataDialog.tsx
@@ -30,9 +30,9 @@ function GetDataDialog({
           Please follow the steps below
           <ol>
             <li>Select at least one dataset</li>
-            <li>Download the cohort participant machine results files using the button below</li>
+            <li>Download the cohort results for machines using the button below</li>
             <li>Change directory to the location of the downloaded files</li>
-            <li>Copy and run the following command</li>
+            <li>Copy and run the command below</li>
           </ol>
         </DialogContentText>
         <div className="mb-4 flex justify-center">
--- a/src/components/GetDataDialog.tsx
+++ b/src/components/GetDataDialog.tsx
@@ -57,7 +57,7 @@ function GetDataDialog({
         </DialogContentText>
       </DialogContent>
       <DialogActions>
-        <Button onClick={() => onClose()} autoFocus>
+        <Button data-cy="get-data-dialog-close-button" onClick={() => onClose()} autoFocus>
           Close
         </Button>
       </DialogActions>
--- a/src/components/GetDataDialog.tsx
+++ b/src/components/GetDataDialog.tsx
@@ -6,8 +6,19 @@ import DialogContentText from '@mui/material/DialogContentText';
 import DialogTitle from '@mui/material/DialogTitle';
 import useMediaQuery from '@mui/material/useMediaQuery';
 import { useTheme } from '@mui/material/styles';
+import DownloadResultButton from './DownloadResultButton';
 
-function GetDataDialog({ open, onClose }: { open: boolean; onClose: () => void }) {
+function GetDataDialog({
+  open,
+  onClose,
+  disableDownloadResultsButton,
+  handleDownloadResultButtonClick,
+}: {
+  open: boolean;
+  onClose: () => void;
+  disableDownloadResultsButton: boolean;
+  handleDownloadResultButtonClick: (identifier: string) => void;
+}) {
   const theme = useTheme();
   const fullScreen = useMediaQuery(theme.breakpoints.down('md'));
 
@@ -19,10 +30,19 @@ function GetDataDialog({ open, onClose }: { open: boolean; onClose: () => void }
           Please follow the steps below
           <ol>
             <li>Select at least one dataset</li>
-            <li>Download the participant-level and dataset-level results files</li>
+            <li>Download the cohort participant machine results files using the button below</li>
             <li>Change directory to the location of the downloaded files</li>
-            <li>Copy and run the following command:</li>
+            <li>Copy and run the following command</li>
           </ol>
+        </DialogContentText>
+        <div className="mb-4 flex justify-center">
+          <DownloadResultButton
+            identifier="cohort participant machine"
+            disabled={disableDownloadResultsButton}
+            handleClick={(identifier) => handleDownloadResultButtonClick(identifier)}
+          />
+        </div>
+        <DialogContentText>
           <code className="text-black">
             docker run -t -v $(pwd):/data neurobagel/dataget:latest /data/dataset-level-results.tsv
             /data/participant-level-results.tsv /data/output
--- a/src/components/GetDataDialog.tsx
+++ b/src/components/GetDataDialog.tsx
@@ -37,7 +37,7 @@ function GetDataDialog({
         </DialogContentText>
         <div className="mb-4 flex justify-center">
           <DownloadResultButton
-            identifier="cohort participant machine"
+            identifier="cohort-participant-machine"
             disabled={disableDownloadResultsButton}
             handleClick={(identifier) => handleDownloadResultButtonClick(identifier)}
           />
--- a/src/components/ResultContainer.tsx
+++ b/src/components/ResultContainer.tsx
@@ -82,11 +82,12 @@ function ResultContainer({
 
     switch (type) {
       case 'sex': {
-        const entry = Object.entries(sexes).find(([, term_IRI]) => {
+        const matchedSex = Object.entries(sexes).find(([, term_IRI]) => {
           const [, uniqueIdentifier] = term_IRI.split(':');
           return uri.includes(uniqueIdentifier);
         });
-        return entry ? entry[0] : uri;
+        // If a match is found, return the label, otherwise return the original URI
+        return matchedSex ? matchedSex[0] : uri;
       }
 
       case 'sessionType':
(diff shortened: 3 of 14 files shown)


src/components/GetDataDialog.tsx
```diff
@@ -19,10 +30,19 @@ function GetDataDialog({ open, onClose }: { open: boolean; onClose: () => void }
           Please follow the steps below
           <ol>
             <li>Select at least one dataset</li>
-            <li>Download the participant-level and dataset-level results files</li>
+            <li>Download the cohort participant machine results files using the button below</li>
             <li>Change directory to the location of the downloaded files</li>
-            <li>Copy and run the following command:</li>
+            <li>Copy and run the following command</li>
```
surchs:
```suggestion
            <li>Copy and run the command below</li>
```

src/components/GetDataDialog.tsx
```diff
@@ -19,10 +30,19 @@ function GetDataDialog({ open, onClose }: { open: boolean; onClose: () => void }
           Please follow the steps below
           <ol>
             <li>Select at least one dataset</li>
-            <li>Download the participant-level and dataset-level results files</li>
+            <li>Download the cohort participant machine results files using the button below</li>
```
surchs:
```suggestion
            <li>Download the cohort results for machines using the button below</li>
```

src/components/GetDataDialog.tsx
```diff
@@ -19,10 +30,19 @@ function GetDataDialog({ open, onClose }: { open: boolean; onClose: () => void }
...
             <li>Change directory to the location of the downloaded files</li>
-            <li>Copy and run the following command:</li>
+            <li>Copy and run the following command</li>
           </ol>
+        </DialogContentText>
+        <div className="mb-4 flex justify-center">
+          <DownloadResultButton
+            identifier="cohort participant machine"
```
surchs: :cherries: 
```suggestion
            identifier="btn-cohort-participant-machine"
```

src/components/ResultContainer.tsx
```diff
@@ -1,11 +1,20 @@
 import { useState, useEffect, useCallback } from 'react';
 import { Button, FormControlLabel, Checkbox, Typography } from '@mui/material';
 import ResultCard from './ResultCard';
-import { QueryResponse, Pipelines } from '../utils/types';
+import { QueryResponse, Pipelines, AttributeOption } from '../utils/types';
```
surchs, line 4: did we just miss this before?

src/components/ResultContainer.tsx
```diff
@@ -58,103 +67,218 @@ function ResultContainer({ response }: { response: QueryResponse | null }) {
...
+      return uri.map((singleUri) => convertURIToLabel(type, singleUri)).join(', ');
+    }
+
+    if (!uri) {
+      return uri;
+    }
+
+    switch (type) {
+      case 'sex': {
+        const entry = Object.entries(sexes).find(([, value]) => {
```
surchs:
```suggestion
        const [label, term_IRI] = Object.entries(sexes).find(([, term_IRI]) => {
```
1. This is complicated stuff, let's use clear and descriptive variable names!
2. Let's use destructuring on the `entry` here directly so I don't have to remember that the key is the label.

src/components/ResultContainer.tsx
```diff
@@ -58,103 +67,218 @@ function ResultContainer({ response }: { response: QueryResponse | null }) {
...
+          const [, id] = value.split(':');
```
surchs:
```suggestion
          const [, unique_identifier] = term_IRI.split(':');
```

src/components/ResultContainer.tsx
```diff
@@ -58,103 +67,218 @@ function ResultContainer({ response }: { response: QueryResponse | null }) {
...
+          return uri.includes(id);
+        });
+        return entry ? entry[0] : uri;
```
surchs:
```suggestion
        return label ? label : uri;
```
see above, I find the `entry[0]` pretty hard to follow

src/components/ResultContainer.tsx
```diff
@@ -58,103 +67,218 @@ function ResultContainer({ response }: { response: QueryResponse | null }) {
...
+        return assessment ? assessment.Label : uri;
+      }
+
+      case 'modality': {
+        const modalityKey = Object.keys(modalities).find((key) => key === uri);
+        return modalityKey ? modalities[modalityKey].label : uri;
+      }
+
+      case 'pipeline':
+        return uri.slice(65);
```
surchs: :eyes:, I mean, no?
```suggestion
        const segments = uri.split('/');
        return segments[segments.length - 1];
```
or anything that is not tied to a fixed string length.

src/components/ResultContainer.tsx
```diff
@@ -58,103 +67,218 @@ function ResultContainer({ response }: { response: QueryResponse | null }) {
...
+      : '';
   }
 
   function generateTSVString(buttonIdentifier: string) {
     if (response) {
       const tsvRows = [];
       const datasets = response.responses.filter((res) => download.includes(res.dataset_uuid));
 
-      if (buttonIdentifier === 'participant-level') {
+      if (buttonIdentifier === 'cohort participant') {
```
surchs: why no `-` anymore?

surchs (changes requested):
Thanks @rmanaem for putting this together so quickly and beautifully. The UI looks really awesome!

I left a couple of clarity comments for you to look at, mostly minor, but some of the new label matching functionality might need another look.

I think your e2e test of the generated TSV fails because during the e2e test you don't actually have a live n-API to talk to, and so the list of diagnoses or attributes with their labels is not "gotten". 

I think we can either mock these requests and then just assert over a smaller set of diagnoses being correctly relabeled - or we go the long route and have a full on e2e test with live n-API and f-API. 

I strongly prefer mocking because these kinds of expensive e2e tests should probably live in a different test suite.

src/components/GetDataDialog.tsx
rmanaem: The issue with this is, the `btn` part and the `-` end up on the button in the UI because we're using the identifier for that. The use of `-` used to work because the words `participant-level` and `dataset-level` were already hyphenated.

src/components/ResultContainer.tsx
rmanaem: No we just didn't needed it in this component.

src/components/ResultContainer.tsx
rmanaem: I mean that's how we're truncating it for the pipeline button but I'm good with this implementation as long as we stay consistent.

src/components/ResultContainer.tsx
rmanaem: See above comment.

src/components/ResultContainer.tsx
rmanaem: Thanks to TS, I was reminded that we can't destruct on entry since `find` may return `undefined` and cause an error.

src/components/ResultContainer.tsx
surchs: OK I see with the Null response type, how about a comment above then to say entry[0] is the label? Not sure there is a way to name this variable better to make it clear without comment. Maybe `matched_pipeline`? But still the same question of what is [0]

src/components/GetDataDialog.tsx
surchs: Ah, I see. I also misunderstood what this identifier does, I thought it was a cypress selector. Let's leave it! And if we ever get confused about the spaces, we can decouple identifier from button text then.

surchs (approved):
Thanks @rmanaem, looks good!

I left another comment on the label:URI disassembler function, let's make it easy for ourselves to get back into the code in a couple weeks with a comment.

:cook:

surchs:
Oh and I suggest `[ENH]` for the PR title because it's user facing.

src/components/ResultContainer.tsx
```diff
@@ -168,7 +294,7 @@ function ResultContainer({ response }: { response: QueryResponse | null }) {
     const element = document.createElement('a');
     const encodedTSV = encodeURIComponent(generateTSVString(buttonIdentifier));
     element.setAttribute('href', `data:text/tab-separated-values;charset=utf-8,${encodedTSV}`);
-    element.setAttribute('download', `${buttonIdentifier}-results.tsv`);
+    element.setAttribute('download', `${buttonIdentifier} results.tsv`);
```
surchs: Hey @rmanaem, I missed this bit. Just realized when looking at your re-generated query tool output files. This means that by default, the downloaded file is called `cohort participant machine results.tsv` - with spaces in it, correct? I really think that's a recipe for things to go bad. For example in Bash you'd have to escape these spaces when pasting the file name to `dataget` - which is the main point of the machine file. So let's make the button text and downloaded file text independent of the button identifier!

surchs (changes requested):
Hey @rmanaem, I missed that our downloaded files now also have spaces in them by default because button identifier, button text, and file name are all the same. 

Spaces in the file names **are** going to be more of an issue than in the identifier, so let's fix that before merging.

Specifically make downloaded file names not have any spaces

src/components/DownloadResultButton.tsx
```diff
@@ -18,7 +18,7 @@ function DownloadResultButton({
       onClick={() => handleClick(identifier)}
       disabled={disabled}
     >
-      Download {identifier} result
+      Download {identifier.split('-').join(' ')} result
```
surchs, line 21: :tada:

surchs (approved):
very cool, thanks @rmanaem 

:cook:

PR #390, opened 2024-12-04
[FIX] Replaced the latest version tag with a baked-in version

### LINKED ISSUE #291: Replace the latest version tag with a baked-in version
If an older version of the tool is deployed, the current implementation still uses the latest release tag. This needs to be replaced with the baked-in version.

 https://github.com/neurobagel/query-tool/blob/0ccfd0ac71bf5c1cca17c3ee7491283ced2df228/src/components/Navbar.tsx#L38-L49

TODO:

- [ ] Figure out how to obtain the version number **during** the Docker image build process (e.g. from GH)
- [ ] Make sure the obtained version number is of the **new release**, i.e. if build happens as part of release, make sure the number is of the new/future release, not the previous one
- [ ] For nightly builds: make sure the version number is nightly
- [ ] Also include a commit sha in the version so we can differentiate nightly deployments
- [ ] Bake this information into the docker image so that every time we deploy a tool from this image, it'll display the correct version number
- [ ] Write a test that checks that the version number is something we can now control directly (e.g. ENV var, hard-coded, whatever it ends up being)
- [ ] Manually test this by triggering a new build (e.g. nightly) and then deploying off of the docker image and confirming the version number

Note:
- preview deployments during PRs via netlify will probably not work like this
- also local dev environments / bare metal installs may not have access to the same information / context so we need to decide what they should show as versions. Ideally something very noticeable so if we ever see it in prod, we know something is wrong

rmanaem:
- Closes #291

- The version is now read from the `package.json`
- Docker workflows have been update:
  - build docker on release:
    - update the version in `package.json` to match that of GitHub
    - commits the change of version in `package.json`
    - trigger the deploy workflow
  - build docker nightly: 
    - update the version in `package.json` to the latest SHA
    - don't commit the change in version

## Checklist

- [x] Tests pass
- [x] Checks pass

For new features:
- [ ] Tests have been added

For bug fixes:

## Summary by Sourcery

Replace the dynamic fetching of the latest release tag with a static version from package.json and update Docker workflows to manage versioning during builds.

Bug Fixes:
- Replace the dynamic fetching of the latest release tag with a static version read from package.json in the Navbar component.

CI:
- Update Docker workflows to handle versioning by reading from package.json and committing changes during release builds.


.github/workflows/build_docker_on_release.yml
```diff
@@ -4,17 +4,47 @@
...
+        id: generate-token
+        uses: actions/create-github-app-token@v1
+        with:
+          app-id: ${{ vars.NB_BOT_ID }}
+          private-key: ${{ secrets.NB_BOT_KEY }}
+
+      - name: Update package.json version
+        run: |
+          # Use jq to update the package.json version
+          jq --arg version "${{ env.version }}" '.version = $version' package.json > temp.json && mv temp.json package.json
```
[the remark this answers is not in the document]
rmanaem: No :)

.github/workflows/build_docker_nightly.yml
```diff
@@ -16,6 +16,19 @@ jobs:
         with:
           submodules: recursive
 
+      - name: Fetch latest commit SHA from main branch
+        id: fetch_sha
+        run: |
+          git fetch origin main
+          LATEST_SHA=$(git rev-parse --short=7 origin/main)
+          echo "LATEST_SHA=$LATEST_SHA" >> $GITHUB_ENV
```
surchs, line 24: fancy stuff but I don't think you need this! Check out the info available in the even body for `push` (and similar for release): https://docs.github.com/en/webhooks/webhook-events-and-payloads#push . I imagine `after` is what you need here

.github/workflows/build_docker_nightly.yml
```diff
@@ -16,6 +16,19 @@ jobs:
...
+
+      - name: Update package.json version
+        run: |
+          SHA=${{ env.LATEST_SHA }}
+          jq ".version = \"build:${SHA}\"" package.json > tmp.$$.json && mv tmp.$$.json package.json
+          cat package.json
```
surchs, line 30: leftover for debug?

.github/workflows/build_docker_nightly.yml
```diff
@@ -16,6 +16,19 @@ jobs:
...
+        id: fetch_sha
+        run: |
+          git fetch origin main
+          LATEST_SHA=$(git rev-parse --short=7 origin/main)
+          echo "LATEST_SHA=$LATEST_SHA" >> $GITHUB_ENV
+
+      - name: Update package.json version
+        run: |
+          SHA=${{ env.LATEST_SHA }}
+          jq ".version = \"build:${SHA}\"" package.json > tmp.$$.json && mv tmp.$$.json package.json
```
surchs, line 29: neato

.github/workflows/build_docker_on_release.yml
```diff
@@ -4,17 +4,47 @@ on:
...
         uses: actions/checkout@v4
         # Some repos have submodules that need to be part of the Docker image
         with:
           submodules: recursive
-      - 
+
+      - name: Extract release version
+        id: extract_version
+        run: |
+          echo "version=${{ github.event.release.tag_name }}" >> $GITHUB_ENV
```
surchs: Not sure I agree here. You save yourself 1 repeated reference to the event payload in exchange for a significantly harder to read workflow. I say we just keep using `github.event.release.tag_name` directly to access this info

.github/workflows/build_docker_on_release.yml
```diff
@@ -4,17 +4,47 @@ on:
...
+
+      - name: Generate a token
+        id: generate-token
+        uses: actions/create-github-app-token@v1
+        with:
+          app-id: ${{ vars.NB_BOT_ID }}
+          private-key: ${{ secrets.NB_BOT_KEY }}
```
surchs: interesting. have we used this before?

.github/workflows/build_docker_on_release.yml
```diff
@@ -4,17 +4,47 @@ on:
...
+
+      - name: Commit updated package.json
+        run: |
+          git config --global user.name "Neurobagel Bot"
+          git config --global user.email "neurobagel-bot[bot]@users.noreply.github.com"
+          git add package.json
+          git commit -m "Update package.json version to match release ${{ env.version }}"
+          git push origin main
+        env:
+          GITHUB_TOKEN: ${{ steps.generate-token.outputs.token }}
```
surchs: I think this is OK, but in an ideal world we would have this commit be part of the release. i.e. if I look at the differ between v0.7.0 and v0.7.1 (https://github.com/neurobagel/query-tool/compare/v0.7.0...v0.7.1) I would see the commit to bump to v0.7.1 in this diff, and not a commit to bump to v0.7.0

I assume the only way to make that happen is for auto to do this somewhere during its business: https://intuit.github.io/auto/docs/plugins/release-lifecycle-hooks

But we can handle that in a later issue when we want to be all fancy

.github/workflows/build_docker_on_release.yml
```diff
@@ -58,3 +88,9 @@ jobs:
             org.opencontainers.image.created=${{ fromJSON(steps.meta.outputs.json).labels['org.opencontainers.image.source'] }}
             org.opencontainers.image.revision=${{ fromJSON(steps.meta.outputs.json).labels['org.opencontainers.image.revision'] }}
             org.opencontainers.image.version=${{ fromJSON(steps.meta.outputs.json).labels['org.opencontainers.image.version'] }}
+
+      - name: Trigger deploy workflow
+        uses: peter-evans/repository-dispatch@v2
+        with:
+          event-type: trigger-deploy
+          token: ${{ steps.generate-token.outputs.token }}
```
surchs: Hhm. This is interesting.  Is this needed because we no longer want to deploy **from** the release commit, but from release commit + 1 (i.e. the package.json version bump)? If so, all the more reason to figure out how to make the version bump part of the release soon. But in the meantime: agreed. But please add a comment to explain why this is needed!

.github/workflows/deploy.yaml
```diff
@@ -2,9 +2,10 @@
 name: deploy
 
 on:
-  release:
-    types: [published]
   workflow_dispatch:
+  repository_dispatch:
+    types:
+      - trigger-deploy
```
surchs: Assuming I understand the need for this new trigger in my comment over in build_on_release, please add a corresponding explanation comment here

src/components/Navbar.tsx
```diff
@@ -17,28 +16,15 @@ import Login from '@mui/icons-material/Login';
 import Avatar from '@mui/material/Avatar';
 import { useAuth0 } from '@auth0/auth0-react';
 import { enableAuth } from '../utils/constants';
+import packageJson from '../../package.json';
```
surchs, line 19: are you sure this path resolves after a full build? i.e. npm run build, not npm run dev?

surchs (approved):
Thanks @rmanaem for this! Looks very nice and solves an important problem :tada: 

I left some comments, mainly clarity. But please also open an issue to explore ways to make the package.json version bump part of the release in the future. I think that'd be clearer 

:cook:

src/components/Navbar.tsx
rmanaem: Yea tested it with the nightly docker image

rmanaem:
After an in-length discussion, we decided to move to implementing a workflow to run on PR when labels have been changed (action type: labeled, see [docs](https://docs.github.com/en/webhooks/webhook-events-and-payloads?actionType=labeled#pull_request)). This workflow will calculate the new version using auto version (see [docs](https://intuit.github.io/auto/docs/generated/version)), update the `package.json`.
The application of `release` label should only happen after the requested changes and comments have been addressed and the PR has been approved to avoid out of sync local and remote.

.github/workflows/build_docker_on_release.yml
rmanaem: Yes pretty much everywhere.

.github/workflows/build_docker_nightly.yml
rmanaem: Yes but also helpful to see at all times

.github/workflows/build_docker_nightly.yml
rmanaem: Yea but that won't work if the event was `workflow_dispatch`

.github/workflows/bump_version.yml
```diff
@@ -0,0 +1,90 @@
...
+            ((MINOR++))
+            PATCH=0
+          elif [ "${{ env.INCREMENT }}" == "patch" ]; then
+            ((PATCH++))
+          fi
+
+          # Construct the new version
+          NEW_VERSION="v$MAJOR.$MINOR.$PATCH"
+          echo "New version: $NEW_VERSION"
+          echo "NEW_VERSION=$NEW_VERSION" >> $GITHUB_ENV
```
surchs, line 76: does this have to happen in two steps? Where else would I use the increment variable

surchs (approved):
Thanks @rmanaem, this looks very good!

Left one question, otherwise :cook:

.github/workflows/bump_version.yml
rmanaem: No, needed separation of concern when testing the script.

PR #434, opened 2025-01-10
[MNT] Automated generation of result examples

### LINKED ISSUE #368: Update auto-generation of query result examples
- Related to neurobagel/neurobagel_examples#45

We need to update the current workflow for auto-generating query result examples, to ensure that the examples are intuitive for a user and represent a typical federated query result structure.

Ideally, we would be able to auto-generate an example that:
- [x] Represents aggregated and non-aggregated dataset results in the _same_ file = one human-readable result TSV example, one machine-optimized result TSV example (this is what we currently have, but the examples were generated manually)

### Steps to implement
- [x] ensure that neurobagel_examples submodule exists in query tool repo
- [x] repurpose our f-API success response example (which contains 2 aggregated datasets and 1 non-aggregated dataset) https://github.com/neurobagel/neurobagel_examples/blob/main/api-responses/fapi_query_success_200.json to mock the f-API response to auto-generate the query result TSVs and to avoid having to stand up two separate nodes in the query tool repo
- [x] create workflow to regenerate query result TSVs using above
  - wf should run:
    - [x] whenever the f-API example outputs in `neurobagel_examples` is updated
    - [x] whenever the query result TSV format itself changes (e.g., columns renamed)

rmanaem:
- Closes #368 

- Implemented e2e test that mocks a [successful fapi response](https://github.com/neurobagel/neurobagel_examples/blob/main/api-responses/fapi_query_success_200.json) to get a mix of aggregated and non aggregated responses in the result files
- Implemented an action to run the e2e test and save the files as artifacts
  - Included a job in the same action to download the artifacts and use them to open a pull request against the `neurobagel_examples` repo

## Checklist

- [x] Tests pass
- [x] Checks pass

For new features:
- [ ] Tests have been added

For bug fixes:

## Summary by Sourcery

Add a workflow to automatically update the query-tool-results example files in the neurobagel_examples repository.

CI:
- Add a new GitHub Actions workflow to generate example result files using end-to-end tests and create a pull request to update them in the neurobagel_examples repository.

Tests:
- Add an end-to-end test to generate example result files by mocking a successful fapi response.

### DIFF
--- a/.github/workflows/component-test.yaml
+++ b/.github/workflows/component-test.yaml
@@ -13,6 +13,8 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v4
+        with:
+          submodules: true
 
       - name: Create .env file
         run: |
--- a/.github/workflows/updated_examples.yaml
+++ b/.github/workflows/updated_examples.yaml
@@ -65,10 +65,6 @@ jobs:
           name: e2e-test-results
           path: neurobagel_examples/query-tool-results
 
-      - name: debug
-        run: |
-          ls
-
       - name: Create Pull Request
         id: create_pr
         uses: peter-evans/create-pull-request@v5
--- a/.github/workflows/updated_examples.yaml
+++ b/.github/workflows/updated_examples.yaml
@@ -10,6 +10,40 @@ on:
 jobs:
   e2e-test:
     runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          submodules: true
+
+      - name: Set up node env
+        uses: actions/setup-node@v4
+        with:
+          node-version: 20
+
+      - name: Create .env file
+        run: |
+          echo -e "NB_API_QUERY_URL=https://federate.neurobagel.org/\nNB_ENABLE_AUTH=true\nNB_QUERY_CLIENT_ID=mockclientid" > .env
+
+      - name: build
+        run: npm install && npm run build
+
+      - name: Run end to end tests
+        uses: cypress-io/github-action@v6
+        with:
+          wait-on: http://localhost:5173
+          start: npm run preview
+          spec: cypress/e2e/UpdateExamples.cy.ts
+          component: false
+
+      - name: Upload test artifacts
+        uses: actions/upload-artifact@v3
+        with:
+          name: e2e-test-results
+          path: cypress/downloads/*
+
+  update-query-tool-results:
+    runs-on: ubuntu-latest
     needs: e2e-test
     steps:
       - name: Generate a token
--- a/.github/workflows/updated_examples.yaml
+++ b/.github/workflows/updated_examples.yaml
@@ -74,6 +74,8 @@ jobs:
 
       - name: Create Pull Request
         run: |
+          git config --global user.name "Neurobagel Bot"
+          git config --global user.email "neurobagel-bot[bot]@users.noreply.github.com"
           git checkout -b update-query-tool-results
           git add .
           git commit -m "Update query-tool-results files"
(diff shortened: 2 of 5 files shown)


.github/workflows/component-test.yaml
```diff
@@ -13,6 +13,8 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v4
+        with:
+          submodules: true
```
surchs, line 17: is this something we had missed before? Not sure I understand why the submodules are needed now but weren't before?

.github/workflows/updated_examples.yaml
```diff
@@ -0,0 +1,77 @@
...
+          wait-on: http://localhost:5173
+          start: npm run preview
+          spec: cypress/e2e/UpdateExamples.cy.ts
+          component: false
+
+      - name: Upload test artifacts
+        uses: actions/upload-artifact@v4
+        with:
+          name: query-tool-results
+          path: cypress/downloads/*
```
surchs, line 40: I wonder if we could just skip this whole job and instead make the existing e2e test workflow upload this artefact.  There is a nice mechanism for this available: https://github.com/github/docs/blob/main/content/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows.md#using-data-from-the-triggering-workflow

But `cypress/e2e/UpdateExamples.cy.ts` isn't a test, but just a way to use cypress to make the example files. So maybe it's more confusing to use the e2e test outputs. In either case, that would be a change for a future PR

cypress.config.ts
```diff
@@ -3,11 +3,13 @@ import { defineConfig } from 'cypress';
 
 export default defineConfig({
   e2e: {
+    specPattern: 'cypress/e2e/*',
```
surchs, line 6: out of interest: what does this do and why is it needed?

cypress/e2e/UpdateExamples.cy.ts
```diff
@@ -0,0 +1,18 @@
+import fapiQuerySuccess200 from '../../neurobagel_examples/api-responses/fapi_query_success_200.json';
+
+describe('Update Examples', () => {
+  it('Removes a newline character from a dataset name in the downloaded dataset-level results file', () => {
```
surchs, line 4:
```suggestion
  it('Generates a new dataset-level results file and removes newline characters', () => {
```
This test / testfile also definitely needs a top-level comment explaining why it exists and why it is used in this odd way (i.e. looks like a test, has no assertions, has a funny name ...)

surchs (approved):
Thanks @rmanaem for fixing this! Very cool workflow.

I left one idea comment, and one request to change before merging (about comment in the test). Take a look, but I'll approve now already

:cook:

.github/workflows/component-test.yaml
rmanaem: This was needed after the new test was added since it uses the fapi example from the neurobagel_examples, the build fails due to a typescript error for the missing module.

.github/workflows/updated_examples.yaml
rmanaem: I see, the main reason I kept it separate from the other workflow is because we generate the result files with different content in other tests and if I'm not mistaken cypress overwrites them so depending on which test runs first we might end up with the wrong examples in the neurobage_examples repo.

cypress/e2e/UpdateExamples.cy.ts
rmanaem: Good catch forgot to update that after I copied the skeleton from another test.
Will add a comment for clarity.

cypress.config.ts
rmanaem: It explicitly tells cypress where to look for e2e test specs and we're using it to prevent the compatibility test from recipes from sneaking their way into the query tool test runs.

PR #450, opened 2025-01-24
[ENH] Move to Docker multistage builds

rmanaem:
- Closes #433 

Changes proposed in this pull request:

- Added `ENV` variables with placeholder values to the Dockerfile 
  - these values are meant to be replaced every time a container starts up
- Added entrypoint script that replaces placeholder values
  - entrypoint won't let container start unless required .env variables are provided

Other note: plan is to release from this PR

**NOTE: If this pull request is to be released, the release label must be applied once the review process is done to avoid the local and remote from going out of sync as a consequence of the `bump version` workflow run**

## Checklist
_This section is for the PR reviewer_

- [x] PR has an interpretable title with a prefix (`[ENH]`, `[FIX]`, `[REF]`, `[TST]`, `[CI]`, `[MNT]`, `[INF]`, `[MODEL]`, `[DOC]`) _(see our [Contributing Guidelines](https://neurobagel.org/contributing/CONTRIBUTING#pull-request-guidelines) for more info)_
- [ ] PR has a label for the release changelog or `skip-release` (to be applied by maintainers only)
- [x] PR links to GitHub issue with mention `Closes #XXXX`
- [x] Tests pass
- [x] Checks pass
- [ ] If the PR changes the participant-level and/or the dataset-level result file, the [`query-tool-results` files](https://github.com/neurobagel/neurobagel_examples/tree/main/query-tool-results) in the  [neurobagel_examples repo](https://github.com/neurobagel/neurobagel_examples/) have been regenerated

For new features:
- [ ] Tests have been added

For bug fixes:
- [ ] There is at least one test that would fail under the original bug conditions.

## Summary by Sourcery

Switch the base image to `node:20-alpine` and copy the built files to an `nginx:alpine` image. Inject a custom plausible link to the docker container at runtime using environment variables.

Build:
- Use a multi-stage build to reduce the final image size.

Deployment:
- Add an entrypoint script to inject a custom plausible link if the `PLAUSIBLE_DOMAIN` and `PLAUSIBLE_SRC` environment variables are set.

## Summary by Sourcery

Update the Dockerfile to use a multi-stage build, switching to `node:20-alpine` for building and `nginx:alpine` for serving. Implement dynamic injection of environment variables at runtime via an entrypoint script, enabling customization of plausible link, authentication settings, and other configurations.

Build:
- Use a multi-stage build to reduce the final image size.

Deployment:
- Add an entrypoint script to inject environment variables at runtime.

PR #468, opened 2025-01-29
[DOC] Document NB_QUERY_HEADER_SCRIPT

surchs:
- Related to https://github.com/neurobagel/query-tool/pull/450

- short description on the new env variable for header scripts

## Checklist

- [ ] PR links to GitHub issue with mention `Closes #XXXX`
- [x] Tests pass
- [x] Checks pass

For new features:
- [ ] Tests have been added

For bug fixes:

## Summary by Sourcery

Documentation:
- Document the `NB_QUERY_HEADER_SCRIPT` environment variable, which allows users to add custom scripts to the header of the query tool.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,7 +45,7 @@ but before proceeding with either you need to set the environment variables.
 | `NB_API_QUERY_URL`       | string  | Yes                                      | -                        | `https://federate.neurobagel.org/`                        |
 | `NB_QUERY_APP_BASE_PATH` | string  | No                                       | `/`                      | `/query/`                                                 |
 | `NB_ENABLE_AUTH`         | boolean | No                                       | `false`                  | `false`                                                   |
-| `NB_QUERY_CLIENT_ID`     | string  | Yes (if `NB_ENABLE_AUTH` is set to true) | -                        | `46923719231972-dhsahgasl3123.apps.googleusercontent.com` |
+| `NB_QUERY_CLIENT_ID`     | string  | Yes (if `NB_ENABLE_AUTH` is set to true) | -                        | `""`                                                      |
 | `NB_QUERY_HEADER_SCRIPT` | string  | No                                       | `""`                     | `'<script defer data-domain="mydomain" src="plausible">'` |
 
 #### `NB_API_QUERY_URL`
--- a/README.md
+++ b/README.md
@@ -70,6 +70,11 @@ _At the moment, query tool uses Google for authentication, so you need to obtain
 If you want to add a custom script to the header of the query tool, you can set the `NB_QUERY_HEADER_SCRIPT` environment variable to the script you want to add.
 This script will be added to the header of the query tool. For example, in our production deployment we use the GDPR aware analytics tool [Plausible](https://plausible.io/),
 so we set the `NB_QUERY_HEADER_SCRIPT` to the script provided by Plausible.
+When you use this variable, make sure you use single quotes to include the script like so:
+
+```bash
+NB_QUERY_HEADER_SCRIPT='<script defer data-domain="mydomain" src="plausible">'
+```
 
 #### Set the environment variables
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -46,6 +46,7 @@ but before proceeding with either you need to set the environment variables.
 | `NB_QUERY_APP_BASE_PATH` | string  | No                                       | `/`                      | `/query/`                                                 |
 | `NB_ENABLE_AUTH`         | boolean | No                                       | `false`                  | `false`                                                   |
 | `NB_QUERY_CLIENT_ID`     | string  | Yes (if `NB_ENABLE_AUTH` is set to true) | -                        | `46923719231972-dhsahgasl3123.apps.googleusercontent.com` |
```
alyssadai: 🍒: I just noticed that the multi-stage PR https://github.com/neurobagel/query-tool/pull/450 changed the default value of `NB_QUERY_CLIENT_ID` to `testclient1234` (I think), which no longer matches the above table:
https://github.com/neurobagel/query-tool/blob/a8f8ce35fa3cf15ce93081d4c3269c0407612ac4/entrypoint.sh#L21

Not sure if this makes sense to revert in this PR, but I'd strongly vote to keep the default value for this variable empty
1. since that's what we have in our docs in the [environment variable table](https://neurobagel.org/user_guide/config/#environment-variables)
2. to avoid confusion about the behaviour of `NB_ENABLE_AUTH` (`NB_QUERY_CLIENT_ID` is only required when `NB_ENABLE_AUTH` is true, and yet the current implemented defaults are contradictory: `NB_ENABLE_AUTH=false` but `NB_QUERY_CLIENT_ID` is non-empty).

README.md
```diff
@@ -64,6 +65,12 @@ If the API you'd like to send queries to requires authentication, you need to se
 If the `NB_ENABLE_AUTH` is set to `true` (it is `false` by default), you need to provide a valid client ID for the authentication.
 _At the moment, query tool uses Google for authentication, so you need to obtain a client ID from Google developer console. See [documentation](https://developers.google.com/identity/gsi/web/guides/get-google-api-clientid) for more information._
 
+#### `NB_QUERY_HEADER_SCRIPT`
+
+If you want to add a custom script to the header of the query tool, you can set the `NB_QUERY_HEADER_SCRIPT` environment variable to the script you want to add.
+This script will be added to the header of the query tool. For example, in our production deployment we use the GDPR aware analytics tool [Plausible](https://plausible.io/),
+so we set the `NB_QUERY_HEADER_SCRIPT` to the script provided by Plausible.
```
alyssadai, line 72: 🍒:
Wondering if we need to provide any guidelines on multi-line scripts? Or would that be obvious do you think?

alyssadai (approved):
Thanks @surchs for the quick fix! I left a comment about another variable ### I realized we changed the default of in https://github.com/neurobagel/query-tool/pull/450 - not sure if want to fix that here, but since they're both README-relevant I thought I'd mention it.

See if you agree, then 🧑‍🍳

(Also, don't forget a PR label!)

README.md
alyssadai: Actually @surchs, we may also want to caution users about quoting. The Plausible script at least includes double quotes  `""` inside, as well as spaces. In this case, I _think_ the whole thing would have to be single quoted in the .env file to be safe, to avoid running into similar issues as we did with passwords?

Not sure how complicated this gets for longer scripts, if mixed quoting is a possibility (?). If so, we may have to take the same approach as for passwords 😬

README.md
surchs: Agree, I'll change it to emtpy

README.md
surchs: yes, we can provide multiple scripts in a single line. And no, I don't think we want to document this all explicitly. I think this variable is mainly for us to put plausible in the public query tool, so I don't see this as a key feature of the tool that we put a lot of effort in for usability - i.e. I wouldn't test a bunch of use cases to make sure they keep working

PR #487, opened 2025-02-14
[FIX] Prevent previously selected pipeline version from appearing in the query after the field has been disabled

### LINKED ISSUE #486: Removing a derivative query parameter disables the "version" field but leaves a previous version tag in the query

### Expected Behavior

If I select a pipeline and version from the dropdown like so, I want to find results that match both pipeline and version! 

Image
If I then reset the pipeline field back to empty, I want to find all results, regardless of pipeline

### Current Behavior

After deselecting the pipeline field, the version tag becomes inaccessible but stays "selected"

Image

The version tag is then also included in requests to the f-API when I submit a query: 
```
https://federate.neurobagel.org/query?node_url=https%3A%2F%2Fqpn.neurobagel.org%2F&pipeline_version=23.1.3
```

### Error message

There is no error message, the results are (likely) wrong. I am not entirely sure what happens if I don't define a pipeline name but I do define a version tag. It seems that at the moment doing so is a valid query. It probably shouldn't and we should fix that in the api repos. 

### Environment

_No response_

### How to reproduce

1. select pipeline and version
2. deselect pipeline
3. run query
4. look at results

### Anything else?

_No response_

rmanaem:
- Closes #486 

- Added a condition in the `constructQueryURL` function to set `pipelineVersion` to null when no `pipelineName` is selected so it doesn't show up in the query
- Set the `inputValue` prop on the disabled to `null` so it is cleared when disabled
- Added tests for both

## Checklist

- [x] Tests pass
- [x] Checks pass

For new features:
- [ ] Tests have been added

For bug fixes:
- [x] There is at least one test that would fail under the original bug conditions.

## Summary by Sourcery

Clear the pipeline version when the pipeline name is deselected.  This prevents including the pipeline version in the query when no pipeline is selected.

Bug Fixes:
- Prevent pipeline version from appearing in the query after the pipeline name field is cleared or disabled

Tests:
- Add tests to verify the fix prevents the pipeline version from appearing in the query when the pipeline name is cleared

### DIFF
--- a/cypress/e2e/APIRequests.cy.ts
+++ b/cypress/e2e/APIRequests.cy.ts
@@ -400,13 +400,21 @@ describe('Successful API query requests', () => {
     cy.get('[data-cy="Maximum age-continuous-field"]').type('30');
     cy.get('[data-cy="Minimum number of imaging sessions-continuous-field"]').type('2');
     cy.get('[data-cy="Minimum number of phenotypic sessions-continuous-field"]').type('3');
+    // Assert that pipeline version doesn't sneak into the query if pipeline name is cleared
+    cy.get('[data-cy="Pipeline name-categorical-field"]').type('fmri{downarrow}{enter}');
+    cy.get('[data-cy="Pipeline version-categorical-field"]').type('0.2.3{downarrow}{enter}');
+    cy.get('[data-cy="Pipeline version-categorical-field"] input').should('have.value', '0.2.3');
+    cy.get('[data-cy="Pipeline name-categorical-field"]').clear();
+    cy.get('[data-cy="Pipeline version-categorical-field"] input').should('be.disabled');
+    cy.get('[data-cy="Pipeline version-categorical-field"]').should('have.value', '');
     cy.get('[data-cy="submit-query-button"]').click();
     cy.wait('@call')
       .its('request.url')
       .should('contain', 'min_age=10')
       .and('contain', 'max_age=30')
       .and('contain', 'min_num_imaging_sessions=2')
-      .and('contain', 'min_num_phenotypic_sessions=3');
+      .and('contain', 'min_num_phenotypic_sessions=3')
+      .and('not.contain', 'pipeline_version');
   });
 });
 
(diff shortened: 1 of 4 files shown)


cypress/e2e/APIRequests.cy.ts
```diff
@@ -400,13 +400,21 @@ describe('Successful API query requests', () => {
...
+    cy.get('[data-cy="Pipeline version-categorical-field"]').should('have.value', '');
     cy.get('[data-cy="submit-query-button"]').click();
     cy.wait('@call')
       .its('request.url')
       .should('contain', 'min_age=10')
       .and('contain', 'max_age=30')
       .and('contain', 'min_num_imaging_sessions=2')
-      .and('contain', 'min_num_phenotypic_sessions=3');
+      .and('contain', 'min_num_phenotypic_sessions=3')
+      .and('not.contain', 'pipeline_version');
```
surchs, line 417: Nice

cypress/e2e/Form.cy.ts
```diff
@@ -85,6 +85,9 @@ describe('App', () => {
     cy.get('[data-cy="Pipeline version-categorical-field"] input').should('not.be.disabled');
     cy.get('[data-cy="Pipeline version-categorical-field"]').type('0.2.3{downarrow}{enter}');
     cy.get('[data-cy="Pipeline version-categorical-field"] input').should('have.value', '0.2.3');
+    cy.get('[data-cy="Pipeline name-categorical-field"]').clear();
+    cy.get('[data-cy="Pipeline version-categorical-field"] input').should('be.disabled');
+    cy.get('[data-cy="Pipeline version-categorical-field"]').should('have.value', '');
```
surchs, line 105: this works, but I think it would be better as a separate test to keep it one test, one problem.

surchs (approved):
Looks good @rmanaem . Got a comment on the test, otherwise :cook: 

sidenote: preview doesn't work well right now because it uses a production f-API as a target. This is another thing a staging environment could help with: be the target for preview deployments

cypress/e2e/Form.cy.ts
rmanaem: like a separate test case i.e., `it` or a different test file?

surchs: I'd say `it` is completely enough

PR #504, opened 2025-03-12
[MNT] Update badge links in README.md

alyssadai:
- Part of https://github.com/neurobagel/planning/issues/235

## Checklist

- [x] Tests pass
- [x] Checks pass

For new features:
- [ ] Tests have been added

For bug fixes:

## Summary by Sourcery

Chores:
- Updates badge links in README.md.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,16 +2,16 @@
 
 # Query Tool
 
-![GitHub branch check runs](https://img.shields.io/github/check-runs/neurobagel/query-tool/main?style=flat-square)
-![GitHub Actions Workflow Status](https://img.shields.io/github/actions/workflow/status/neurobagel/query-tool/e2e-test.yaml?branch=main&style=flat-square&label=e2e%20tests&link=https%3A%2F%2Fgithub.com%2Fneurobagel%2Fquery-tool%2Factions%2Fworkflows%2Fe2e-test.yaml)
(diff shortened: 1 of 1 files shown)


rmanaem (commented):
Thanks for the PR @alyssadai!
Just have two questions:

- The Docker version badge and Docker pulls badge both point to the query-tool on Docker hub, is that intentional? Should we point Docker version to Docker itself like we do with node?
- Do you know why the badge for checks is failing even though everything is green?
image

alyssadai:
Hey @rmanaem,

> The Docker version badge and Docker pulls badge both point to the query-tool on Docker hub, is that intentional? Should we point Docker version to Docker itself like we do with node?

Yes, it's intentional, since they both refer to a Docker image for our tool rather than the version of a dependency. I think we want both links to take the user to the tags for our tools so they can pull it, etc. - they wouldn't be doing this directly with something like Node or Python (which we don't expect them to install directly).

> Do you know why the badge for checks is failing even though everything is green?

I'm not sure tbh...is it possible the status is calculated based on whether any workflows that run on main have failed in their most recent run? If so, I see that Dependabot updates was apparently failing: https://github.com/neurobagel/query-tool/actions/workflows/dependabot/dependabot-updates?query=branch%3Amain

Not sure if that makes sense. @surchs, any ideas?

rmanaem:
> I'm not sure tbh...is it possible the status is calculated based on whether any workflows that run on main have failed 
in their most recent run? If so, I see that Dependabot updates was apparently failing:

Yea, that could be it.

surchs:
fwiw, the checks badge is also showing failing right now.

alyssadai:
I'll merge this PR in that case - if the failing status persists where we expect things to be passing, we can investigate that separately.

alyssadai:
Oops, just realized it's missing an approval still 🙂

surchs (approved):
🧑‍🍳

PR #628, opened 2025-09-15
[FIX] Fixed the `Cannot select "All"` bug

### LINKED ISSUE #134: [BUG] Cannot select "All" nodes directly via the UI

### Expected Behavior

- a specific node is selected
- I'd rather search "All" nodes
- I see "All" in the dropdown for nodes
- I click on "All" in the node list

-> Now "All" should be selected

### Current Behavior

tool_no_work_good

Clicking "All" in the node list doesn't do anything.
I can only get "All" by deselecting the last node.

### Error message

user sad

### Environment

_No response_

### How to reproduce

see snippet

### Anything else?

_No response_

rmanaem:
- Closes #134 

## Checklist

- [x] Tests pass
- [x] Checks pass

For new features:
- [ ] Tests have been added

For bug fixes:

## Summary by Sourcery

Fix the nodes field selection logic to ensure the “All” option overrides other selections and defaults when cleared, and add Cypress intercept and e2e tests to validate this behavior

Bug Fixes:
- Ensure selecting “All” in the nodes field overrides any other selected options and defaults to “All” when cleared

Enhancements:
- Refactor nodes field handler to separately manage “All” selection and strip it when other options are chosen

Tests:
- Add Cypress intercept for `/nodes` and include it in setup waits
- Add new E2E test to cover nodes field selection scenarios, including selecting and clearing “All” and other options

### DIFF
--- a/cypress/e2e/Form.cy.ts
+++ b/cypress/e2e/Form.cy.ts
@@ -111,7 +111,7 @@ describe('App', () => {
     cy.get('[data-cy="filter-toggle-button"]').should('not.exist');
     cy.get('[data-cy="query-form-container"]').should('be.visible');
   });
-  it('Selects different values for nodes field', () => {
+  it('Selects different nodes in the nodes field', () => {
     cy.get('[data-cy="Neurobagel graph-categorical-field"] input').type(
       'OpenNeur{downarrow}{enter}'
     );
--- a/cypress/e2e/Form.cy.ts
+++ b/cypress/e2e/Form.cy.ts
@@ -1,4 +1,5 @@
 import {
+  nodeOptions,
   diagnosisOptions,
   pipelineOptions,
   pipelineVersionOptions,
@@ -6,6 +7,13 @@ import {
 
 describe('App', () => {
   beforeEach(() => {
+    cy.intercept(
+      {
+        method: 'GET',
+        url: '/nodes',
+      },
+      nodeOptions
+    ).as('getNodes');
     cy.intercept(
       {
         method: 'GET',
@@ -21,7 +29,7 @@ describe('App', () => {
       pipelineOptions
     ).as('getPipelineOptions');
     cy.visit('/');
-    cy.wait(['@getDiagnosisOptions', '@getPipelineOptions']);
+    cy.wait(['@getNodes', '@getDiagnosisOptions', '@getPipelineOptions']);
 
     // TODO: remove this
     // Bit of a hacky way to close the auth dialog
@@ -102,5 +110,29 @@ describe('App', () => {
     cy.viewport(1200, 800); // Desktop viewport
     cy.get('[data-cy="filter-toggle-button"]').should('not.exist');
     cy.get('[data-cy="query-form-container"]').should('be.visible');
+  });
+  it('Selects different values for nodes field', () => {
+    cy.get('[data-cy="Neurobagel graph-categorical-field"] input').type(
+      'OpenNeur{downarrow}{enter}'
+    );
+    cy.get('[data-cy="Neurobagel graph-categorical-field"]').should('contain', 'OpenNeuro');
+    cy.get('[data-cy="Neurobagel graph-categorical-field"] input').type('Quebec{downarrow}{enter}');
+    cy.get('[data-cy="Neurobagel graph-categorical-field"]')
+      .should('contain', 'Quebec')
+      .and('contain', 'OpenNeuro');
+    cy.get('[data-cy="Neurobagel graph-categorical-field"] input').type('All{downarrow}{enter}');
+    cy.get('[data-cy="Neurobagel graph-categorical-field"]')
+      .should('not.contain', 'Quebec')
+      .and('not.contain', 'OpenNeuro');
+    cy.get('[data-cy="Neurobagel graph-categorical-field"]').should('contain', 'All');
+    cy.get('[data-cy="Neurobagel graph-categorical-field"] input').type(
+      'OpenNeur{downarrow}{enter}'
+    );
+    cy.get('[data-cy="Neurobagel graph-categorical-field"] input').type('Quebec{downarrow}{enter}');
+    cy.get('.MuiAutocomplete-clearIndicator').click();
+    cy.get('[data-cy="Neurobagel graph-categorical-field"]')
+      .should('not.contain', 'Quebec')
+      .and('not.contain', 'OpenNeuro');
+    cy.get('[data-cy="Neurobagel graph-categorical-field"]').should('contain', 'All');
   });
 });
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -305,6 +305,9 @@ function App() {
   function updateCategoricalQueryParams(fieldLabel: string, value: FieldInput) {
     switch (fieldLabel) {
       case 'Neurobagel graph':
+        // Cases are to identify whether "All" was
+        // 1. explicitly selected (by the user) i.e., "All" is at the end of the array
+        // 2. selected by default on startup or when the field is cleared i.e., "All" is the only option in the array
         if (Array.isArray(value)) {
           // When “All” is selected after some other nodes were already selected, strip all other options
           if (value.length > 1 && value[value.length - 1].label === 'All') {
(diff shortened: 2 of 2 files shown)


cypress/e2e/Form.cy.ts
```diff
@@ -103,4 +111,28 @@ describe('App', () => {
     cy.get('[data-cy="filter-toggle-button"]').should('not.exist');
     cy.get('[data-cy="query-form-container"]').should('be.visible');
   });
+  it('Selects different values for nodes field', () => {
```
surchs:
```suggestion
  it('Selects different nodes in the nodes field', () => {
```

src/App.tsx
```diff
@@ -306,15 +306,21 @@ function App() {
     switch (fieldLabel) {
       case 'Neurobagel graph':
         if (Array.isArray(value)) {
-          // If no option is selected default to All
+          // When “All” is selected, override any other options
+          if (value.length > 1 && value[value.length - 1].label === 'All') {
```
surchs, line 312: why does it have to be the most recent entry? Shouldn't `any` be used here?

src/App.tsx
```diff
@@ -306,15 +306,21 @@ function App() {
...
-          } else if (value.length > 1) {
-            setSearchParams({ node: value.filter((n) => n.label !== 'All').map((n) => n.label) });
-          } else {
-            setSearchParams({ node: value.map((n) => n.label) });
+            break;
           }
+
+          // strip "All" from the list if there are other options selected
+          const withoutAll = value.filter((n) => n.label !== 'All').map((n) => n.label);
+          setSearchParams({ node: withoutAll });
```
surchs: I'm not sure what this means. How would I be able to have `All` selected alongside other options. Isn't that already handled by your change earlier in this script.

src/App.tsx
surchs: That should also remove the need for your special handle below for stripping `All` again if other values are present

surchs (changes requested):
Thanks @rmanaem. I think this all makes sense / works well, but we should clarify the implementation in a couple spots. I left some comments there

src/App.tsx
rmanaem: > why does it have to be the most recent entry? Shouldn't any be used here?

It has to be the most recent entry because that is the last selection by the user. Essentially we want to wipe everything except "All" when "All" is selected after some other nodes were already selected.

> That should also remove the need for your special handle below for stripping All again if other values are present

We can't condition "All" to go away when some other node is selected and for it to wipe other nodes when "All" is selected. We have to address them separately.

src/App.tsx
rmanaem: This is to address the initial node selection since the default value for the dropdown is "All" and also when we wipe the selection using the dismiss button (`x`) on the dropdown we go back to default. So when a user first makes selection other than the default we need to remove "All" manually.

rmanaem:
Thanks for the review @surchs!
I've responded to your comments and updated the in-line comments. Let me know if that clarifies the implementation.

src/App.tsx
surchs: > It has to be the most recent entry because that is the last selection by the user. Essentially we want to wipe everything except "All" when "All" is selected after some other nodes were already selected.

I don't understand your explanation. iiuc: if All is anywhere in `value`,  then we want to "override" `value` and make it only be `All`. Tying this condition to your implementation knowledge that `"All"` should only ever be the last entry in the array because of how it is added is what I was referring to here. So unless I misunderstand, `any` is a better condition here and also semantically closer to what you are checking.

> We can't condition "All" to go away when some other node is selected and for it to wipe other nodes when "All" is selected. We have to address them separately.

Also don't understand the explanation. iiuc the flow goes:

- is more than one node selected? -> is any of them `All`? -> then remove all nodes except `All` -> exit
- is no node selected? -> then add node `All` -> exit
- if neither of above, remove `All` from the nodes list -> exit

That last point to me should never be relevant, if we can remove "All" in point 1

src/App.tsx
```diff
@@ -305,16 +305,25 @@ function App() {
   function updateCategoricalQueryParams(fieldLabel: string, value: FieldInput) {
     switch (fieldLabel) {
       case 'Neurobagel graph':
+        // Cases are to identify whether "All" was
+        // 1. explicitly selected (by the user) i.e., "All" is at the end of the array
+        // 2. selected by default on startup or when the field is cleared i.e., "All" is the only option in the array
```
surchs:
```suggestion
        // 2. "All" is the only option in the array, i.e. "All" was selected by default on startup or after the field is cleared
        // 3. No node selected, e.g. when the app start, then we select "All" as the default value
        // 4. A user has selected a specific node (i.e. not "All"), then "All" (as the default value) is still in the array, but at the front - and we need to remove "All"
```

src/App.tsx
```diff
@@ -305,16 +305,25 @@ function App() {
...
             setSearchParams({ node: ['All'] });
-            // If any option beside All is selected, remove All
-          } else if (value.length > 1) {
-            setSearchParams({ node: value.filter((n) => n.label !== 'All').map((n) => n.label) });
-          } else {
-            setSearchParams({ node: value.map((n) => n.label) });
+            break;
           }
+
+          // Since "All" is the default, we need to strip "All" from the list when any other option is selected
```
surchs:
```suggestion
          // Case 4: If a user added a specific node (not "All"), we need to remove "All" (the default value) from the array 
```

src/App.tsx
```diff
@@ -305,16 +305,25 @@ function App() {
   function updateCategoricalQueryParams(fieldLabel: string, value: FieldInput) {
     switch (fieldLabel) {
       case 'Neurobagel graph':
+        // Cases are to identify whether "All" was
+        // 1. explicitly selected (by the user) i.e., "All" is at the end of the array
+        // 2. selected by default on startup or when the field is cleared i.e., "All" is the only option in the array
         if (Array.isArray(value)) {
-          // If no option is selected default to All
+          // When “All” is selected after some other nodes were already selected, strip all other options
```
surchs:
```suggestion
          // Case 1: When “All” is selected after some other nodes were already selected, strip all other options
```

src/App.tsx
```diff
@@ -305,16 +305,25 @@ function App() {
...
+          if (value.length > 1 && value[value.length - 1].label === 'All') {
+            setSearchParams({ node: ['All'] });
+            break;
+          }
+
+          // default to “All” when nodes are cleared
```
surchs:
```suggestion
          // Case 3: default to “All” when nodes are cleared
```

src/App.tsx
surchs: OK, I almost get it. But: what happens in case 2? 

- `value.length=1`
- therefore it's not case 1
- and not case 3
- so case 4 catches and removes "All"

do we just re-render then with `value.length=0` and that triggers case `3`, setting "All"?

surchs (approved):
Thanks @rmanaem, this makes sense now.

I have one question left, please take a look.

Then :cook:

rmanaem:
@surchs Could you please take a one last look before we finalize this? 
Also I believe we'll have to wait on releasing this until the APIs have been switched.

src/App.tsx
```diff
@@ -305,23 +305,27 @@ function App() {
...
           if (value.length === 0) {
             setSearchParams({ node: ['All'] });
             break;
           }
 
-          // Since "All" is the default, we need to strip "All" from the list when any other option is selected
+          // Case 3: User selected some nodes, but not "All"
+          // in this case if "All" is present in the selected options array,
+          // "All" would be the first option in the array as the default and fallback option
+          // Approach: Remove "All" if it was in the selected options and keep the rest
```
surchs:
```suggestion
          // Case 3: User selected some nodes, but not the "All" option.
          // If "All" is present in the selected options array, 
          // it is leftover at the start of the array from being initially set as the default, 
          // and we need to remove it.
          // Approach: Remove "All" if it was in the selected options and keep the rest
```

surchs (approved):
Looks great. Very clear. Thanks @rmanaem 

:cook:

PR #651, opened 2025-10-22
[ENH] Split query into dataset- and subject-level query

### LINKED ISSUE #614: Split fetching results logic into a two step process
At the moment, we send one GET request to fetch all the results (both datasets and subjects) from the f-api and we use those to both display result cards and for downloading.
With the introduction of new POST /dataset and /subjects endpoints, the query tool will now have to send a POST request to the /datasets endpoint once the submit query button is clicked and only send a request to the /subjects endpoint when the user clicks a download button.

- [x] Update type definitions
  - type definition for POST /datasets response (dataset info without subject arrays)
  - type definition for POST /datasets request body
  - type definition for POST /subjects response (subject-level data) 
  - type definition for POST /subjects request body (includes query params + dataset UUIDs)
  - update `Result` interface to reflect that `subject_data` won't be present in the dataset queries
- [x] Create `buildQueryRequestBody` function to convert current query params to JSON object
- [x] Update submitQuery
   - replace GET request to /query with POST to /datasets
   - use `buildQueryRequestBody` to create request body
   - update result state to handle new response structure
- [x] Implement subject data fetching
  - create new `fetchSubjectData` function
  - update `downloadResults` function to generate tsv from the fetched subject data
- [x] Update constants
  - add `datasetsURL` constant for POST /datasets endpoint
  - add `subjectsURL` constant for POST /subjects endpoint
- [x] verify protected datasets work correctly with the new flow
- [x] Implement tests
  - update existing tests and mocks for /query endpoint to now use /datasets endpoint
- [x] Add loading state indicator for download button

rmanaem:
- Closes #614

## Checklist

- [x] Tests pass
- [x] Checks pass

For new features:
- [ ] Tests have been added

For bug fixes:

## Summary by Sourcery

Separate dataset and subject fetching into two POST endpoints, refactor request body construction, update download logic to fetch subject data before TSV generation, add error handling and loading states, and adjust types, constants, and tests accordingly.

Enhancements:
- Refactor App to construct and send POST requests to /datasets with a JSON request body instead of GET URL parameters
- Add constructDatasetsRequestBody and constructSubjectsRequestBody helpers to build payloads for dataset and subject endpoints
- Update ResultContainer to fetch subject-level responses on download, handle errors with snackbars, and construct TSVs from subject data
- Enhance DownloadResultButton to support asynchronous downloads with a loading spinner and disabled state
- Change API endpoints in constants to datasetsURL and subjectsURL and switch axios calls accordingly
- Expand types with DatasetsResponse, SubjectsResponse, DatasetsRequestBody, and SubjectsRequestBody definitions

Tests:
- Update Cypress fixtures with dataset and subject response snippets
- Modify intercepts in E2E and component tests to use POST /datasets and /subjects and assert over request bodies
- Add subject response fixtures and adapt ResultsTSV, APIRequests, UpdateExamples, and Checkbox tests to new two-step fetching flow


src/components/ResultContainer.tsx
```diff
@@ -138,51 +155,94 @@
...
+      'DatasetImagingModalities',
+      'DatasetPipelines',
+    ].join('\t');
+    tsvRows.push(headers);
+
+    subjectsData.responses.forEach((res) => {
+      if (res.records_protected) {
+        tsvRows.push(
+          [
+            res.dataset_name.replace('\n', ' '),
```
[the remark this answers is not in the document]
rmanaem: No

src/components/DownloadResultButton.tsx
```diff
@@ -18,11 +19,12 @@ function DownloadResultButton({
   handleClick,
 }: {
   disabled: boolean;
-  handleClick: (index: number) => void;
+  handleClick: (index: number) => Promise<void>;
 }) {
   const [open, setOpen] = useState(false);
   const anchorRef = useRef<HTMLDivElement>(null);
   const [selectedIndex, setSelectedIndex] = useState(0);
+  const [loading, setLoading] = useState(false);
```
surchs: the actual querying and generation of the result .tsv is currently handled in `ResultContainer.tsx` - I'd say that's then also where the responsibility for keeping the loading state should fall. In other words: the "go handle the fetching of results" concerns should be kept together in the same responsible component

surchs: iirc this should then also remove the need for the async wrapper a couple lines below

src/components/DownloadResultButton.tsx
```diff
@@ -41,11 +43,27 @@ function DownloadResultButton({
...
 
+  const handleDownloadClick = async () => {
+    setLoading(true);
+    try {
+      await handleClick(selectedIndex);
+    } finally {
+      setLoading(false);
+    }
+  };
+
```
surchs: should move to ResultContainer, see above comment

src/components/ResultContainer.tsx
```diff
@@ -138,51 +155,94 @@ function ResultContainer({
...
+            isFileWithLabels
+              ? convertURIToLabel(
+                  'pipeline',
+                  parsePipelinesInfoToString(res.available_pipelines).split(',')
+                )
+              : parsePipelinesInfoToString(res.available_pipelines),
+          ].join('\t')
+        );
+      } else {
+        // @ts-expect-error: typescript doesn't know that subject_data is an array when records_protected is false.
```
surchs, line 214: just noticing this here again. This might be a good reason to ensure that our API responses don't change type between aggregation modes. i.e. always return Array type

src/components/ResultContainer.tsx
```diff
@@ -194,72 +254,99 @@ function ResultContainer({
...
-    element.setAttribute('download', fileName);
+  const action = (snackbarId: SnackbarKey) => (
+    <IconButton
+      onClick={() => {
+        closeSnackbar(snackbarId);
+      }}
+    >
+      <CloseIcon className="text-white" />
+    </IconButton>
+  );
```
surchs: wouldn't it make more sense here to use our error area? i.e. not add the snackbar here for actual errors (we're using it for warnings atm).

src/components/ResultContainer.tsx
```diff
@@ -194,72 +254,99 @@ function ResultContainer({
...
+    return tsvRows.join('\n');
+  }
+
+  function constructSubjectsRequestBody(
+    requestBody: DatasetsRequestBody,
+    selectedDatasetUuids: string[],
+    datasetResponses: DatasetsResponse['responses'],
+    nodes: NodeOption[]
+  ): SubjectsRequestBody {
+    // Create a mapping from node_name to node_url (ApiURL)
```
surchs: I don't think it's a good idea to 

a) put the subject query generation inside the ResultContainer when the responsibility for the dataset query is already in App.tsx - those belong together and should be kept in the same context.
b) Make the subject query depend directly on the dataset query. Even if this makes the code currently simpler, it is harder to understand and creates a coupling that we will have to later disentangle again. It probably also makes testing quite a bit harder to do / reason about.

So instead: 

- let's keep dataset and study request body generation in the same place (ResultContainer may be more appropriate than App.tsx for both)
- make both request body generations independent of each other / read the query parameters independently

src/utils/types.ts
```diff
@@ -53,21 +53,54 @@ export interface Subject {
...
+  assessment?: string;
+  image_modal?: string;
+  pipeline_name?: string;
+  pipeline_version?: string;
+  nodes: Array<{ node_url: string }>;
+}
+
+export interface SubjectsRequestBody extends Omit<DatasetsRequestBody, 'nodes'> {
+  nodes: Array<{ node_url: string; dataset_uuids: string[] }>;
 }
```
surchs: As mentioned elsewhere, I don't think it's a good idea to couple these requests. To me, the fact that the request bodies look so similar is a coincidence at the moment - and not something we can count on being true in the future. So I would define their interfaces independently to make that clear, even if that means some duplication

src/App.tsx
```diff
@@ -609,6 +578,9 @@ function App() {
                 response={sortedResults || null}
                 diagnosisOptions={diagnosisOptions}
                 assessmentOptions={assessmentOptions}
+                datasetsRequestBody={result ? constructDatasetsRequestBody() : null}
+                availableNodes={availableNodes}
+                IDToken={IDToken}
               />
```
surchs: also mentioned in ResultContainer.tsx: both request bodies should be built and sent (i.e. the request and response handled) in the same context/component. Whether that is better done in App.tsx or ResultContainer.tsx I am not sure. Might depend on which context currently has more of the necessary information / would require less prop drilling.

cypress/e2e/APIRequests.cy.ts
```diff
@@ -408,13 +408,13 @@ describe('Successful API query requests', () => {
...
-      .and('contain', 'min_num_phenotypic_sessions=3')
-      .and('not.contain', 'pipeline_version');
+    cy.wait('@call').then((interception) => {
+      expect(interception.request.body).to.have.property('min_age', 10);
+      expect(interception.request.body).to.have.property('max_age', 30);
+      expect(interception.request.body).to.have.property('min_num_imaging_sessions', 2);
+      expect(interception.request.body).to.have.property('min_num_phenotypic_sessions', 3);
+      expect(interception.request.body).to.not.have.property('pipeline_version');
+    });
   });
```
surchs, line 418: :tada:

cypress/component/DownloadResultButton.cy.tsx
```diff
@@ -2,7 +2,7 @@ import DownloadResultButton from '../../src/components/DownloadResultButton';
 
 const props = {
   disabled: false,
-  handleClick: () => {},
+  handleClick: async () => {},
```
surchs: You should add a test to ensure the new loading state is actually correctly shown. Could be as simple as:

if loading=true -> button grey and spinner there. (that's assuming loading state is lifted out of the component and to the parent who is already handling the response)

surchs (changes requested):
Thanks @rmanaem - app behaves very nicely and the UX is good. I have two kinds of requests:

simple:
- we should make the logic for both request handlers (make request body, send request, parse result, make tsv) live in the same space / component. i.e.: the (old) dataset request and the (new) subject request should be together
- the requests should be independent (not coupled) and not depend on current similarities in the shapes - those may/should be able to change

more conceptual:
- Because we now have two steps (1. find datasets that match cohort, 2. retrieve subject info that match cohort in selected datasets), we should explicitly keep track of this thing `"Cohort"` - also so that a user can't start changing the "cohort" parameters in between running the dataset query (step 1) and downloading the subject info (step 2). At the moment you are doing this implicitly by passing the response object from the dataset query (step 1) to the function that generates the subject request (step 2), but as mentioned, that couples both together. Less importantly, it also does not make this two-step distinction clear in the UX.
- I think we should discuss / consider how we want to deal with this in the UX. One option is to trigger a (dataset) query immediately when a user makes any change to the QueryForm (would require fast queries). Another is to have a flow where we make it clear to the user that they "design" a cohort and that this cohort gets "locked" when they hit the submit button. And once the results (dataset) for this cohort come back, they can explore (i.e. scroll and select) and maybe download subsets of the cohort, but they can't make changes to the cohort criteria (i.e. the QueryForm is locked) without creating a completely new cohort. 

I realize this may be a bigger change, and we definitely should discuss how we want this to behave. But I think it makes sense to do this now so at a minimum we don't have to do it implicitly like you are doing right now.

src/components/ResultContainer.tsx
rmanaem: What do u mean by error area?

surchs: https://github.com/neurobagel/query-tool/blob/73e278db97dbec05b70ed860923b2dec38164aa9/src/components/ErrorAlert.tsx

src/components/ResultContainer.tsx
rmanaem: Not sure I'm following.

cypress/e2e/UndoQueryFormChanges.cy.ts
```diff
@@ -0,0 +1,61 @@
...
+
+    cy.get('[data-cy="close-auth-dialog-button"]').click();
+  });
+
+  function runQueryAndSelectFirstDataset() {
+    cy.get('[data-cy="submit-query-button"]').click();
+    cy.wait('@datasetsQuery');
+    cy.get('[data-cy="card-https://someportal.org/datasets/ds0001-checkbox"] input').check();
+    cy.get('[data-cy="download-results-button"]').should('not.be.disabled');
+  }
```
surchs: If we want to make this reusable, then we should use a custom cypress command: https://docs.cypress.io/api/cypress-api/custom-commands

That's not a bad idea, but out of scope for this PR, I'd say

src/components/ContinuousField.tsx
```diff
@@ -3,27 +3,25 @@ import TextField from '@mui/material/TextField';
...
-  onFieldChange: (fieldLabel: string, value: number | null) => void;
+  value: string;
+  onFieldChange: (fieldLabel: string, value: string) => void;
 }
 
-function ContinuousField({ helperText, label, onFieldChange }: ContinuousFieldProps) {
-  const showError: boolean = helperText !== '';
+function ContinuousField({ helperText, label, value, onFieldChange }: ContinuousFieldProps) {
+  const normalizedHelperText = helperText ?? '';
+  const showError: boolean = normalizedHelperText !== '';
```
surchs: Missed this before but:
- we should probably rename `helperText` to `errorText` given that it is used as a signal for an error
- can we not define a default value for `helperText` and get rid of that nullish coalescing here?

src/components/QueryForm.tsx
```diff
@@ -38,24 +38,34 @@ function QueryForm({
...
-  minAge: number | null;
-  maxAge: number | null;
+  minAge: string;
+  maxAge: string;
   sex: FieldInput;
   diagnosis: FieldInput;
-  minNumImagingSessions: number | null;
-  minNumPhenotypicSessions: number | null;
+  minNumImagingSessions: string;
+  minNumPhenotypicSessions: string;
```
surchs, line 47: I feel like I kinda get why these parameters are all string, but not really. Is it because the components these get passed to are TextFields? Could you explain what has changed here that makes this type change necessary? Especially because this introduces the need for all that string-to-number casting below

src/components/QueryForm.tsx
```diff
@@ -38,24 +38,34 @@ function QueryForm({
...
 
+  function parseNumericValue(value: string): number | null {
+    const trimmedValue = value.trim();
+    if (trimmedValue === '') {
+      return null;
+    }
+
+    const parsed = Number(trimmedValue);
+    return Number.isNaN(parsed) ? Number.NaN : parsed;
+  }
```
surchs: do we have to write this: https://lodash.info/doc/toNumber? And if we do, let's put it in utils and test it

src/components/DownloadResultButton.tsx
```diff
@@ -44,12 +45,7 @@ function DownloadResultButton({
   };
 
   const handleDownloadClick = async () => {
-    setLoading(true);
-    try {
-      await handleClick(selectedIndex);
-    } finally {
-      setLoading(false);
-    }
+    handleClick(selectedIndex);
```
surchs: can't you just pass `handleClick` to the button now? It seems that `handleDownloadClick` is no longer async anyway.

src/components/ResultContainer.tsx
surchs: right now, subject_data can be a string or an array. Because of that, you need to escape a typescript error here (iiuc). I think that's a problem/mistake on the part of the f-API in that it should not change the type of its responses (string vs array) across modes (aggregated vs disaggregated)

src/components/DownloadResultButton.tsx
```diff
@@ -41,11 +44,22 @@ function DownloadResultButton({
     setOpen(false);
   };
 
+  const handleDownloadClick = async () => {
+    handleClick(selectedIndex);
```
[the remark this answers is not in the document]
rmanaem: Good catch

src/components/QueryForm.tsx
```diff
@@ -101,13 +124,15 @@ function QueryForm({
         <ContinuousField
           helperText={minAgeExceedsMaxAge ? '' : minAgeHelperText}
           label="Minimum age"
+          value={minAge}
```
surchs, line 117: What was going on here with the state management before? Did we just get lucky that the Continuous and Categorical input components were showing the right text because they didn't reset between renders? I.e. did we forget to pass `value` back down to them to directly control what they display?

In general: please help me out a bit here with the review by adding a summary to the PR of bugs or issues that you have been finding and drive-by fixing in the PR but that were not in the original issue. I can probably figure it out on my own eventually, but it's just so much faster if you give me some context

src/hooks/useSendQueries.ts
```diff
@@ -0,0 +1,79 @@
...
+}
+
+export default function useSendQueries(IDToken: string | undefined) {
+  async function sendDatasetsQuery(datasetsRequestBody: QueryParams) {
+    const response = await axios.post<DatasetsResponse>(datasetsURL, datasetsRequestBody, {
+      headers: {
+        ...(IDToken ? { Authorization: `Bearer ${IDToken}` } : {}),
+        'Content-Type': 'application/json',
+      },
+    });
```
surchs: This seems a bit overly complex to me. iiuc the only reason you're using this structure here is to get a closure with the ID token. I'd say that's less clear / readable and not necessary. You could pass the token to each function separately or factor out the axios client and header - and then pass the axios instance to the functions. 

The datasets and subject query functions look good to me, and I find it much clearer to import them directly inside the components - rather than to destruct them from this parent-function.

Lastly: I don't think anything inside this file is / needs to be a hook. You don't use other hooks, don't use state, and there is no need to memoize this either. So I would suggest you move the subject and dataset query functions individually to utils.ts

src/App.tsx
```diff
@@ -374,13 +420,23 @@ function App() {
...
+  function parseNumericParam(value: string): number | null {
+    const trimmed = value.trim();
+    if (trimmed === '') {
+      return null;
+    }
+
+    const parsed = Number(trimmed);
+    return Number.isNaN(parsed) ? null : parsed;
+  }
+
```
surchs: that's pretty close to the helper function you have in `QueryForm`. I'd consider looking into whether you can use `null` (or `nan`) on both scenarios, so you don't need to keep two functions around. And as mentioned above I would also suggest using an existing converter like the one from lodash as the basis.

src/App.tsx
```diff
@@ -574,13 +670,28 @@ function App() {
...
+                  action={
+                    <Button color="inherit" size="small" onClick={() => undoQueryFormChanges()}>
+                      Undo changes
+                    </Button>
+                  }
+                  data-cy="query-form-changed-alert"
+                >
+                  You have edited the query form fields since you&apos;ve submitted your query.
+                  Downloading is disabled until you undo the changes or submit the new query.
+                </Alert>
```
surchs, line 660: I think I see why you can't directly reuse `ErrorAlert` but it would be nice if we could converge on a single component for handling these kinds of errors. Otherwise they will just start stacking up. 

Also:

```suggestion
                <Alert
                  severity="error"
                  className="mb-2"
                  action={
                    <Button color="inherit" size="small" onClick={() => undoQueryFormChanges()}>
                      Undo changes
                    </Button>
                  }
                  data-cy="query-form-changed-alert"
                >
                  You have edited the query form fields since you&apos;ve submitted your query.
                  Downloading is disabled until you undo the changes or submit a new query.
                </Alert>
```

cypress/component/ContinuousField.cy.tsx
```diff
@@ -35,8 +43,24 @@ describe('ContinuousField', () => {
...
+          label={defaultProps.label}
+          value={value}
+          onFieldChange={(label, nextValue) => {
+            setValue(nextValue);
+            onFieldChangeSpy(label, nextValue);
+          }}
+        />
+      );
+    }
+    cy.mount(<ControlledField />);
```
surchs: This is clever, but I'd say that's not the point. The component now no longer handles its own state. It (and thus the test) has really only two concerns:

1. If I get passed a `value`, does it show up correctly?
2. If the user enters (or adds to) a value, does the right handler get called with the right payload?

Each of these you can test without simulating external state here, and then we also don't have this (itself untested) piece of logic in our tests. If you really want to also ensure that the component can handle multi-character strings, then you can initialize the test with `value=1` or something like that. Adding `0` should then give you the expected `'10'` payload in the handler.

surchs: And just to be clear: you still need to add the test for case 1

cypress/component/DownloadResultButton.cy.tsx
```diff
@@ -35,14 +48,31 @@ describe('DownloadResultButton', () => {
...
+      <DownloadResultButton
+        disabled={props.disabled}
+        handleClick={handleClickSpy}
+        loading={props.loading}
+      />
+    );
     cy.get('[data-cy="download-results-button"]').click();
     cy.get('@handleClickSpy').should('have.been.calledWith', 0);
   });
+  it('Disables the button and displays a loading spinner when loading is true', () => {
```
surchs, line 67: 🎉

cypress/component/QueryForm.cy.tsx
```diff
@@ -55,25 +56,25 @@ describe('QueryForm', () => {
...
+        minNumPhenotypicSessions={defaultProps.minNumPhenotypicSessions}
+        assessmentTool={defaultProps.assessmentTool}
+        imagingModality={defaultProps.imagingModality}
+        pipelineVersion={defaultProps.pipelineVersion}
+        pipelineName={defaultProps.pipelineName}
+        pipelines={defaultProps.pipelines}
+        updateCategoricalQueryParams={defaultProps.updateCategoricalQueryParams}
+        updateContinuousQueryParams={defaultProps.updateContinuousQueryParams}
+        loading={defaultProps.loading}
+        onSubmitQuery={defaultProps.onSubmitQuery}
```
surchs, line 76: just a thought here but maybe we should consider disabling the no-spreading rule in eslint **for tests only** to make them a bit more readable. wdyt?

cypress/component/QueryForm.cy.tsx
```diff
@@ -155,54 +156,64 @@ describe('QueryForm', () => {
...
+          pipelineVersion={defaultProps.pipelineVersion}
+          pipelineName={defaultProps.pipelineName}
+          pipelines={defaultProps.pipelines}
+          updateCategoricalQueryParams={defaultProps.updateCategoricalQueryParams}
+          updateContinuousQueryParams={(label, value) => {
+            setMinAge(value);
+            updateContinuousQueryParamsSpy(label, value);
+          }}
+          loading={defaultProps.loading}
+          onSubmitQuery={defaultProps.onSubmitQuery}
```
surchs: Same comment here as for the continuousField test: If the component doesn't have state, then keeping state is not its concern anymore and we don't need to test that / simulate it. If the thing you're really concerned over is whether the component can handle the second (or third) character someone enters, then just initialize that state by setting the value explicitly.

surchs (commented):
Hey @rmanaem, thanks for working on this. I haven't gotten to everything yet, but commenting already so you get at least the partial review to get started with.

Main concerns are complexity in the new hook (that I think is a disguised util) and the tests.  The rest looks good.

cypress/component/QueryForm.cy.tsx
rmanaem: I don't mind the current setup but if it helps we can do that in a separate issue.

cypress/e2e/UndoQueryFormChanges.cy.ts
rmanaem: Yea was hoping to reduce duplication but we'll remove for now.

src/components/QueryForm.tsx
rmanaem: So I noticed this as I was working on the `undo changes` functionality. I reckon after we learned about controlled component pattern, we didn't update the `ContinuousField` component and it wasn't fully controlled this whole time. 
iiuc the component managed its own internal state via the `TextField` and the field worked because the TextField's internal state persisted between renders.

> In general: please help me out a bit here with the review by adding a summary to the PR of bugs or issues that you have been finding and drive-by fixing in the PR but that were not in the original issue. I can probably figure it out on my own eventually, but it's just so much faster if you give me some context

Will do in the future.

src/components/QueryForm.tsx
rmanaem: Yea let's stick with the current implementation. I moved it to `utils` and added a test.

src/components/QueryForm.tsx
rmanaem: Yea, `TextField` only works with strings. Changed it to make ContinuousField fully controlled by the parent. Before it was parsing internally which meant we couldn't control the exact display value or preserve invalid input states. The explicit casting is more verbose but keeps all validation logic in one place and makes undo changes feature possible.

src/App.tsx
rmanaem: > I think I see why you can't directly reuse ErrorAlert but it would be nice if we could converge on a single component for handling these kinds of errors. Otherwise they will just start stacking up.

Agreed. We could refactor `ErrorAlert` to handle more functionality.

cypress/component/QueryForm.cy.tsx
surchs: if you don't mind it, then let's leave it until either of minds more. It is more explicit, just a bit repetitive to read

cypress/unit/utils.cy.ts
```diff
@@ -0,0 +1,43 @@
...
+  });
+
+  it('returns null for mixed alphanumeric string', () => {
+    expect(parseNumericValue('12abc')).to.equal(null);
+  });
+
+  it('returns null for special characters', () => {
+    expect(parseNumericValue('!@#')).to.equal(null);
+  });
+});
```
surchs, line 43: I'm confused. We already have unit tests in vitest: https://github.com/neurobagel/annotation-tool/blob/311e0bbebca79063c7a8cb10450ef4f944301ef5/src/utils/util.test.ts

What is different about the tests you have here that makes it so you need to run them in cypress instead? I'd really rather not have 2 parallel unit test frameworks if we can at all avoid it

cypress.config.ts
```diff
@@ -11,7 +11,7 @@ export default defineConfig({
   },
 
   component: {
-    specPattern: 'cypress/component/*',
+    specPattern: ['cypress/component/*', 'cypress/unit/*'],
```
surchs, line 14: related to my other comment on why do we need cypress unit tests in addition to vitest

src/App.tsx
```diff
@@ -110,6 +121,35 @@ function App() {
     </IconButton>
   );
 
+  const currentQueryFormState: QueryFormState = useMemo(
```
surchs: I may have missed this, but what are you memoizing here? I don't see a function call. Isn't this just a normal state variable?

surchs (approved):
Thanks @rmanaem. This all looks good, so I am approving. But before you merge, please take a look at

- cypress unit tests -> why do we need these over vitest
- memoization of what seems to be a static object

And see if you can address those. Then we can merge :cook:

cypress/unit/utils.cy.ts
surchs: turns out, this is a different app

cypress.config.ts
rmanaem: Because we don't have vitest unit tests set up for query tool yet.

PR #656, opened 2025-10-27
[ENH] Add npm plugin to auto

### LINKED ISSUE #588: version tag update inside package.json on release is not working

### Expected behavior

the `version` key in package.json is up to date with the most recent release tag from GitHub after a release. 

### Current behavior

`version` in package.json has not been updated in the last releases:

https://github.com/neurobagel/query-tool/blob/35d55f2ece1a76122e99920bee087648c6c5ad2e/package.json#L4

### Error message

wrong version is shown on a local build

### Environment

- OS:
- Python/Node version:
- Docker version:
- Docker Compose version:

### How to reproduce

- fresh clone and 
- then `npm run dev` 
- look at the version displayed

### Anything else?

We tried to get this to work in https://github.com/neurobagel/query-tool/pull/526 but apparently it does not work. 

Because #658 did not fix the problem, we should:

- [ ] check if we can safely remove / disable `git-tag` and let `npm` make the git tag during release
- [ ] check that the changelog is/will be (since we can't really test) still correctly generated
- [ ] apply the change
- [ ] merge
- [ ] observe if the solution works and then close the issue

surchs:
This should now enable npm to bump the version in package.json during a release.

- Closes #588

- Add `npm` plugin to auto config
- Because we have `private: true`, npm will not publish even if we did configure it: https://docs.npmjs.com/cli/v7/configuring-npm/package-json#private
- auto should take care of all steps, including bumping version in package.json on the next release
- not sure how we would easily test this before a release 

## Checklist

- [ ] PR has an interpretable title with a prefix (`[ENH]`, `[FIX]`, `[REF]`, `[TST]`, `[CI]`, `[MNT]`, `[INF]`, `[MODEL]`, `[DOC]`) _(see our [Contributing Guidelines](https://neurobagel.org/contributing/CONTRIBUTING#pull-request-guidelines) for more info)_
- [ ] PR links to GitHub issue with mention `Closes #XXXX`
- [ ] Tests pass
- [ ] Checks pass

For new features:
- [ ] Tests have been added

For bug fixes:

## Summary by Sourcery

Enhancements:
- Add npm plugin to auto configuration for automatic package.json version bump on release

### DIFF
--- a/.autorc
+++ b/.autorc
@@ -15,6 +15,12 @@
                 "subject": "[pre-commit.ci]",
                 "labels": ["_bot", "skip-release"]
             }
+        ],
+        [
+            "npm",
+            {
+                "setRcToken": false
+            }
         ]
     ],
     "labels": [


alyssadai (approved):
Based on what I can find in the intuit/auto docs about NPM, this looks sound to me 🧑‍🍳

PR #681, opened 2025-12-01
[ENH] Add timestamp to downloaded result filenames

### LINKED ISSUE #568: Add timestamp or unique ID to filename of each results file
If a user downloads results for multiple queries, on certain OSes they may be saved as:
- `neurobagel_query_results.tsv`
- `neurobagel_query_results (1).tsv`
- etc.

This makes it difficult to quickly identify what each file contains. Having a timestamp suffix (e.g., reflecting date & time at download) may help to differentiate results files.

Tusharjamdade:
- Closes #568 

- Added timestamp suffix to downloaded result filenames to prevent OS auto-renaming (e.g., "neurobagel-query-results (1).tsv").
- Filenames now include a clean, sortable timestamp (e.g., "neurobagel-query-results-2025-12-01-23-55-17.tsv").
- Format: YYYY-MM-DD-HH-MM-SS.

## Checklist

- [x] Tests pass
- [x] Checks pass

For new features:
- [ ] Tests have been added

For bug fixes:

## Summary by Sourcery

New Features:
- Append a sortable YYYY-MM-DD-HH-MM-SS timestamp to downloaded query result TSV filenames, including the URI-annotated variant.


src/components/ResultContainer.tsx
```diff
@@ -268,11 +268,16 @@ function ResultContainer({
     try {
       const subjectsResponse = await onDownload(buttonIndex, download);
       setDownloadError(null);
+      const timestamp = new Date()
+        .toISOString()
+        .replace('T', '-')
+        .replace(/:/g, '-')
+        .replace(/\..+/, '');
```
rmanaem: I suggest changing the format of the timestamp to the following: YYYYMMDDHHMMSS for a less-clutered and cleaner file name. Wdyt?

src/components/ResultContainer.tsx
```diff
@@ -268,11 +268,16 @@ function ResultContainer({
...
 
       const fileName =
         buttonIndex === 0
-          ? 'neurobagel-query-results.tsv'
-          : 'neurobagel-query-results-with-URIs.tsv';
+          ? `neurobagel-query-results-${timestamp}.tsv`
+          : `neurobagel-query-results-with-URIs-${timestamp}.tsv`;
```
rmanaem: I suggest using `_` underscore here to better separate the Neurobagel section of the file name from the timestamp. Wdyt?

```suggestion
          ? `neurobagel-query-results_${timestamp}.tsv`
          : `neurobagel-query-results-with-URIs_${timestamp}.tsv`;
```

cypress/e2e/ResultsTSV.cy.ts
```diff
@@ -76,7 +76,7 @@ describe('Results TSV', () => {
     cy.contains('URIs').click();
     cy.get('[data-cy="download-results-button"]').click();
     cy.wait('@subjectsCall');
-    cy.readFile('cypress/downloads/neurobagel-query-results-with-URIs.tsv').should(
+    cy.readFile('cypress/downloads/neurobagel-query-results-with-URIs-*.tsv').should(
```
rmanaem: I believe the `Cy.readFile` doesn't support glob as input so the wild card here is not going to work as we'd like it to: https://docs.cypress.io/api/commands/readfile
My suggestion would be to take a different approach:
Use `Cy.exec` which let's you run a shell command to dynamically grab the filename of the latest file downloaded and then pass it to `Cy.readFile`. For a cleaner approach you could factor it into a helper function and reuse it for all the tests that need it:
```TS
function readLatestFile(pattern: string) {
  return cy.exec(`ls -t ${pattern} | head -n1`).then(({ stdout }) => cy.readFile(stdout.trim()));
}
```
Use in test:
```TS
    readLatestFile('cypress/downloads/neurobagel-query-results-with-URIs-*.tsv').should(
      'contain',
      'some cool name'
    );
```

rmanaem (changes requested):
Hi @Tusharjamdade, thanks for opening the PR. Please see the requested changes below.

src/components/ResultContainer.tsx
Tusharjamdade: Currently, the filename looks like neurobagel-query-results-2025-12-01-18-41-16. I think this format provides good readability.
Should we keep it as is, or switch to a continuous timestamp format like YYYYMMDDHHMMSS by removing the hyphens?

src/components/ResultContainer.tsx
Tusharjamdade: Yes, we can do that!

Tusharjamdade:
Hi @rmanaem,

The tests are now working correctly. The output filenames are generated in the neurobagel-query-results_YYYYMMDDHHMMSS format (e.g., neurobagel-query-results_20251211162958), as per your suggestion.

Please let me know if you would recommend any further improvements

cypress/e2e/ResultsTSV.cy.ts
```diff
@@ -10,6 +10,20 @@ import {
...
+    const file = stdout.trim();
+
+    if (!file) {
+      throw new Error(`No file found for pattern: ${pattern}`);
+    }
+
+    return cy.readFile(file);
+  });
+}
+
```
rmanaem, line 23: I'm assuming you had to go with this implementation because `cy.exec` runs command in the system's default shell and in Windows `ls` and `head` don't exist by default. If that's the case, I suggest using `cy.task` instead and sticking with Node APIs (so no shell commands) for an OS-agnostic implementation.

rmanaem (changes requested):
Hi @Tusharjamdade,
Thanks for implementing the requested changes! I left a comment regarding the test please take a look and let me know what you think.

Tusharjamdade:
Hi @rmanaem,
I’ve updated the tests. Please let me know if there are any changes or improvements needed.

rmanaem (approved):
Thanks @Tusharjamdade I believe this is good to go 🧑‍🍳

PR #684, opened 2025-12-05
[ENH] Fetch imaging modality options from an API endpoint

### LINKED ISSUE #643: Fetch imaging modality options from an API endpoint instead of a using hardcoded list
Having run into the #642 we realized it might no longer be viable to store imaging modality options as a hard coded list on the query tool side and instead they should be retrieved from an API endpoint just like diagnosis and assessment tool options.

Since we will soon have more modalities, we could switch to color-coding them by the BIDS datatype (`anat`, `func`, etc.) instead  of assigning a distinct color per suffix. These datatypes should be able to be read from a config file - see https://github.com/neurobagel/communities/issues/43.

Tasks:
- [x]  Fetch imaging modality options from the API (`nb:Image`) the same way diagnoses/assessments are retrieved, replacing the hardcoded constants.
- [x] Store modality metadata (Label, TermURL, abbreviation, data_type) and use the metadata to map result data to UI labels.
- [x] Generate a distinct color for modality based on `DataType` 
  - [x] Filter out modalities that miss `DataType` or `Abbreviation`
- [x] Update `convertURIToLabel` conversion logic to use fetched modality metadata instead of the static map.
- [x] Update affected component and e2e tests 
- [x] Add tests for the new API request endpoint

References:
- BIDS datatypes (15 total): https://github.com/bids-standard/bids-specification/blob/master/src/schema/objects/datatypes.yaml

rmanaem:
- Closes #643 

- Added imaging modality-specific types (`ImagingModalityOption`, `ImagingModalitiesMetadata`) and a generic `AttributeResponse` 
- Extended the shared attribute fetch helper in `App` to pull the imaging-modalities endpoint with DataType/Abbreviation casing.
- Built the imaging modality metadata map from the API with both prefixed TermURL and full IRI keys to resemble the old hard-coded `modalities` structure
- Introduced a data type to color mapping to replace the old hardcoded modality list in `constants`
- Updated `ResultContainer` and `ResultCard` to consume API-driven modality metadata for labels/abbreviations/colors and to skip modalities missing DataType or Abbreviation
- Expanded mocks with imaging modality success/empty/partial/failure responses and added intercepts for /imaging-modalities in relevant component and e2e tests
- Added assertions for imaging modality notifications (info/warning/error)`APIRequests` e2e suite
- Ensured modality chips render from API metadata in `ResultCard` and `ResultContainer` component tests

## Checklist

- [x] Tests pass
- [x] Checks pass

For new features:
- [ ] Tests have been added

For bug fixes:

## Summary by Sourcery

Fetch imaging modality options and metadata from a new API endpoint instead of hardcoded constants, and propagate this data through the query form and results views.

New Features:
- Load imaging modality options and metadata from an /imaging-modalities API endpoint and surface them in the query form for filtering.
- Provide API-driven imaging modality labels, abbreviations, and colors in result cards and containers using a data type–to–color mapping.

Enhancements:
- Generalize the shared attribute-fetching helper to support typed responses, configurable label filtering, and imaging modality–specific notifications.
- Replace the static imaging modalities constant with a data-type-to-color mapping and metadata lookup keyed by both prefixed and full IRIs.

Tests:
- Extend Cypress e2e and component tests with mocks and assertions for imaging modality API responses, notifications, and rendered modality chips in the results and TSV flows.


src/utils/constants.ts
```diff
@@ -30,55 +30,20 @@ export const sexes: { [key: string]: string } = {
...
-  },
-  'http://purl.org/nidash/nidm#PositronEmissionTomography': {
-    label: 'Positron Emission Tomography',
-    TermURL: 'nidm:PositronEmissionTomography',
-    name: 'PET',
-    bgColor: 'rgb(78,20,186)',
-  },
+export const modalitiesDataTypeColorMapping: { [key: string]: string } = {
+  anat: 'rgb(113, 113, 122)',
+  beh: 'rgb(205, 92, 92)',
```
surchs: We're not supporting `beh` as a BIDS tag

src/utils/constants.ts
```diff
@@ -30,55 +30,20 @@ export const sexes: { [key: string]: string } = {
...
+  fmap: 'rgb(189, 183, 107)',
+  func: 'rgb(143, 188, 143)',
+  ieeg: 'rgb(215,145,50)',
+  meg: 'rgb(78,20,186)',
+  micr: 'rgba(0, 192, 6, 0.54)',
+  motion: 'rgb(199, 21, 133)',
+  mrs: 'rgb(0, 191, 255)',
+  perf: 'rgb(34, 139, 34)',
+  pet: 'rgb(255, 69, 0)',
+  phenotype: 'rgb(138, 43, 226)',
```
surchs: Let's not map this either. This is tabular data about a participant, BIDS handles it, but we don't consider this "imaging"

src/App.tsx
```diff
@@ -137,81 +142,105 @@ function App() {
   };
 
   useEffect(() => {
-    async function getAttributes(NBResource: string, dataElementURI: string) {
+    // T is constrained to include an optional Label so we can safely filter out missing labels when needed.
```
surchs: I don't understand what this comment means. Can you explain it please

src/App.tsx
```diff
@@ -137,81 +142,105 @@ function App() {
...
+      'pipelines',
+      'nb:Pipeline',
+      (pipelineResponse) => {
         pipelineResponse.forEach((option) => {
           setPipelines((prevPipelines) => ({ ...prevPipelines, [option.TermURL]: [] }));
         });
-      }
-    });
+      },
+      { removeMissingLabels: false }
```
surchs, line 238: maybe add a comment here to explain why pipelines have no label / need this special mode

surchs: Could also make the comment a TODO to revisit once pipelines come with Label in the hopefully near future

src/App.tsx
```diff
@@ -137,81 +142,105 @@ function App() {
...
-        return null;
+        enqueueSnackbar(`Failed to retrieve ${NBResource} options`, { variant: 'error', action });
       }
     }
 
-    getAttributes('diagnoses', 'nb:Diagnosis').then((diagnosisResponse) => {
-      if (diagnosisResponse !== null && diagnosisResponse.length !== 0) {
-        setDiagnosisOptions(diagnosisResponse);
-      }
+    getAttributes<AttributeOption>('diagnoses', 'nb:Diagnosis', (diagnosisResponse) => {
```
surchs, line 209: Why remove the guards for diagnoses being null or empty? That seems like a good check

src/App.tsx
```diff
@@ -137,81 +142,105 @@ function App() {
...
-      if (assessmentResponse !== null && assessmentResponse.length !== 0) {
-        setAssessmentOptions(assessmentResponse);
-      }
+    getAttributes<AttributeOption>('assessments', 'nb:Assessment', (assessmentResponse) => {
+      setAssessmentOptions(assessmentResponse);
     });
 
-    getAttributes('pipelines', 'nb:Pipeline').then((pipelineResponse) => {
-      if (pipelineResponse !== null && pipelineResponse.length !== 0) {
+    getAttributes<ImagingModalityOption>('imaging-modalities', 'nb:Image', (modalities) => {
```
surchs, line 217: What should happen if imaging modalities are empty (as is the case in the preview build because we're testing against a f-API that doesn't have these endpoints yet)

surchs: I.e. shouldn't that also produce a warning in the warning area?

src/App.tsx
```diff
@@ -137,81 +142,105 @@ function App() {
...
+          ]);
+          return;
+        }
+
+        let filteredItems = items;
+        const itemsWithLabels = items as Array<{ Label?: string | null }>;
+        if (
+          removeMissingLabels &&
+          itemsWithLabels.some(
+            (item) => Object.prototype.hasOwnProperty.call(item, 'Label') && item.Label === null
```
surchs: I think https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwn replaces hasOwnProperty since quite a while. Maybe bot still recommends though

surchs: Also: what happens here if `item` does not have `"Label"` at all? The first part of your `&&` is false, so `item` would not count as a "missing label", right?

surchs (approved):
Thanks @rmanaem. It seems you have inconsistent type requirements between `AttributeOption` (requires Label) and `getAttributes` (optional Label / can be null). 

Could you take a look at that before merging please. also left a couple of other comments.

Sidenote: preview does not work in this case because it's testing against fAPI that does not return the expected imaging data. This is a bigger problem but would be good to come up with some ideas for how to address this (i.e.  preview build against staging rather than production)

I'm approving with the above changes
🧑‍🍳

src/App.tsx
rmanaem: Of course, this is just meant to provide quick explainer for the generic type `T` constraint we're applying on the `getAttribute` function i.e., whatever type you call it with must optionally have a `Label` prop so we can safely check that label is not null when removing missing labels. Without it TS complains that `Label` isn't a prop of T.

src/App.tsx
rmanaem: > I think https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwn replaces hasOwnProperty since quite a while. Maybe bot still recommends though

I did a quick check and apparently `Object.hasOwn` is part of the ES2022 JS spec and we're running the ES2020 which still doesn't include the new shorthand.

src/App.tsx
rmanaem: I believe we still do that see line 178-187. We also have a test that covers the behaviour.

src/App.tsx
rmanaem: Same as above for diagnosis and assessment, we emit an info toast and have a test for coverage.

src/utils/constants.ts
rmanaem: I believe this was included in the datatypes.yaml link from the issue: https://github.com/bids-standard/bids-specification/blob/master/src/schema/objects/datatypes.yaml
If you recall we decided to include tags we don't support yet.

src/utils/constants.ts
rmanaem: See above.

src/App.tsx
surchs: Aha, good call. I remember seeing this in the "early days", but indeed those are not long ago for me.

src/components/ResultContainer.tsx
```diff
@@ -131,8 +134,11 @@ function ResultContainer({
       }
 
       case 'modality': {
-        const modalityKey = Object.keys(modalities).find((key) => key === uri);
-        return modalityKey ? modalities[modalityKey].label : uri;
+        const modality = Object.values(imagingModalitiesMetadata).find((m) => {
+          const [, id] = m.TermURL.split(':');
+          return uri.includes(id);
+        });
```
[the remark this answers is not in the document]
surchs: @rmanaem, I didn't look deeper into this. Do we have both full URIs and also prefixed URIs?

PR #697, opened 2026-01-10
[ENH] Display enriched dataset level info on result cards

### LINKED ISSUE #682: Fetch dataset level information from f-API and process it for display

### Summary

As a QT user, I want to see dataset level information for the datasets that are displayed to me (e.g. access level, authors, available assessments + imaging modalities ...), so that I can make an informed decision about the usefulness of the dataset before I take additional steps of accessing the data (e.g. requesting or downloading it).

TODO
- [x] Fetch dataset level information from the f-API (or another external source) about the datasets I am discovering with my cohort queries
- [x] Process the received (external) dataset level information internally so it can display it to me
- [x] Create new / update existing UI components for the different types of data that can be retrieved from the external source
  - [x] Update types to adhere to the updated  dataset results coming from f-api
  - [x] Update mocked responses to adhere to the updated dataset and subject results
  - [x] Update `ResultCard` component to show the new fields
  - [x] Update `ResultContainer` component to merge the subject results and the dataset results for creating the output tsv
  - [x] Update `ResultCard` component test to assert over the new UI

It may make sense to split the component / UI work (mocking data) from the fetching / processing work (mocking API or direct integration with API). For components, we should lean towards small, reusable/composable UI components rather than big, hard-coded components.

### Documentation update(s) needed

- [x] neurobagel/documentation#355

rmanaem:
- Closes #682 

- Enriched `ResultCard` UI to display full dataset metadata (authors, access info, contact, repo URL) with robust fallback states.
- Updated ResultContainer to merge dataset metadata into TSV exports, ensuring `RepositoryURI` and `AccessLink` are always populated.
- Aligned `DatasetsResult` and `SubjectsResult` interfaces in `types` with the latest API spec and synced all mock fixtures.
- Updated `ResultCard` component tests for verifying metadata display and handling of missing optional fields.

## Checklist

- [x] Tests pass
- [x] Checks pass

For new features:
- [x] Tests have been added

For bug fixes:

## Summary by Sourcery

Enrich dataset result cards and TSV exports with expanded dataset-level metadata and updated API-aligned types.

New Features:
- Display detailed dataset metadata (authors, access info, repository links, keywords, and references) on result cards with expandable sections and robust fallbacks.
- Expose dataset access level, node mode, raw data availability, and derivative pipeline badges directly in the result card UI.
- Include repository URLs and dataset access links in subject-level TSV exports, merging dataset metadata into protected and unprotected rows.

Enhancements:
- Align datasets and subjects result types and fixtures with the latest API schema, including new metadata fields and renamed properties.
- Refine imaging modality color mapping for anatomical data and adjust result card layout for clearer presentation.

Tests:
- Extend ResultCard Cypress tests to cover new metadata UI, fallback states, and updated pipelines/modality rendering.
- Update ResultContainer Cypress tests and fixtures to reflect the new response shapes, property names, and TSV output fields.


README.md
```diff
@@ -197,9 +197,9 @@ To define a cohort, set your inclusion criteria using the following:
 Once you've defined your criteria, submit them as a query and the query tool will display the results.\
 The query tool offers two different TSV files for results:
 
-- cohort participant results TSV contains: dataset name, portal uri, number of matching subjects, subject id, session id, session file path, session type, age, sex, diagnosis, assessment, number of matching phenotypic sessions, number of matching imaging sessions, session imaging modality, session completed pipelines, dataset imaging modality, and dataset pipelines
+- cohort participant results TSV contains: dataset name, repository url, number of matching subjects, subject id, session id, session file path, session type, age, sex, diagnosis, assessment, number of matching phenotypic sessions, number of matching imaging sessions, session imaging modality, session completed pipelines, dataset imaging modality, and dataset pipelines
```
surchs: just sanity check: did we not *add* a new column as well?

cypress/component/ResultCard.cy.tsx
```diff
@@ -31,85 +32,225 @@ const props = {
...
+      referencesAndLinks: [],
+      keywords: [],
+      repositoryUrl: null,
+      accessInstructions: null,
+      accessType: null,
+      accessEmail: null,
+      accessLink: null,
+      imageModals: [],
+      availablePipelines: {},
+    };
```
surchs, line 121: This is smart - but would it not be clearer to just feed the empty values to the props directly? Especially since you don't do prop spreading when feeding them to the component?

cypress/e2e/UpdateExamples.cy.ts
```diff
@@ -15,16 +16,41 @@ describe('Update Examples', () => {
...
     cy.get('[data-cy="select-all-checkbox"]').find('input').check();
+    /* 
+    Explicitly wait for the downloaded files to exist to ensure
+    the test fails if the download fails silently (e.g., due to a data mismatch).
+    */
+
+    const checkForFile = (pattern: string) => {
+      // Retry logic: cy.task yields the result. If null, retry recursively until timeout.
+      const timeout = 10000;
+      const start = Date.now();
```
surchs: I believe this is misleading or possibly wrong. Because cypress is interpreted synchronously but then executed asynchronously, the `start` time might be quite a bit before the test actually executes.

cypress/e2e/UpdateExamples.cy.ts
```diff
@@ -15,16 +16,41 @@ describe('Update Examples', () => {
...
+
+      const check = () => {
+        cy.task('getLatestFile', pattern).then((found) => {
```
surchs: related to above: afaik cypress already has the ability to auto-retry (https://docs.cypress.io/app/core-concepts/retry-ability). So rather than re-implement it here, we should understand how to use the cypress re-try. Seems to me that maybe all it takes is to add an assertion like `.should('contain', 'something')` https://docs.cypress.io/app/core-concepts/retry-ability#Only-queries-are-retried 
Could you give that a shot?

cypress/e2e/UpdateExamples.cy.ts
```diff
@@ -15,16 +16,41 @@ describe('Update Examples', () => {
...
+          if (!found) {
+            if (Date.now() - start > timeout) {
+              throw new Error(`File matching ${pattern} not found after ${timeout}ms`);
+            }
+            check();
```
surchs: again, be careful with synchronous code in cypress tests: https://docs.cypress.io/app/core-concepts/introduction-to-cypress#Mixing-Async-and-Sync-code

I am not sure what your `check()` call here does exactly, but most likely it'll keep running when cypress **parses** the test, not when it **runs** the test. So you may end up making a long queue of cypress calls here which is probably not what you want

src/App.tsx
```diff
@@ -351,29 +351,39 @@ function App() {
...
       case 'Neurobagel graph':
         if (Array.isArray(value)) {
+          const isAllSelectedNow = value.some((n) => n.label === 'All');
+          const wasAllSelectedBefore = Array.isArray(selectedNode)
+            ? selectedNode.some((n) => n.label === 'All')
+            : false;
+
           // Case 1: User explicitly selected "All" after some other nodes were already selected.
-          // in this case "All" would be the last option in the array
           // Approach: keep "All", remove all other options
```
surchs, line 360: this is neat, but let's generally do these things in a dedicated PR to keep the diff about one thing. For the next PR

src/components/ResultCard.tsx
```diff
@@ -1,169 +1,525 @@
...
+    authors: string[];
+    homepage: string | null;
+    referencesAndLinks: string[];
+    keywords: string[];
+    repositoryUrl: string | null;
+    accessInstructions: string | null;
+    accessType: 'public' | 'registered' | 'restricted' | null;
+    accessEmail: string | null;
+    accessLink: string | null;
     datasetTotalSubjects: number;
```
surchs: this is getting a bit wild with the number of props. One thing we could do is group related info into objects (define interface up top) and then pass these objects. That'd also help keep things a bit more understandable. e.g. groups could be
- datasetInfo
- accessInfo
- nodeInfo
or something like that, each with their subfields.

Another possibility is that - because you are just passing the datasetResponse object down via props, you instead pass the entire datasetResponse object as a single prop. semantically a bit less clear than several prop groups, but still better than 21 props

src/components/ResultCard.tsx
```diff
@@ -1,169 +1,525 @@
...
+                  fontSize="small"
+                  data-cy={`card-${datasetUuid}-${type}-access-type-icon`}
+                />
+              );
+            })}
+          </div>
+        </Tooltip>
+      );
+    };
+
```
surchs: Why do you make these inline functions here rather than creating actual components? this file is getting very long, hard to read, and has a ton of concerns. I'd say let's factor out these components - especially if you clearly identify them as you do here.

src/components/ResultCard.tsx
```diff
@@ -1,169 +1,525 @@
...
+          endIcon={<OpenInNewIcon />}
+        >
+          Repository
+        </Button>
+      );
+    };
+
+    return (
+      <Card data-cy={`card-${datasetUuid}`} sx={{ mb: 2 }}>
+        {/* HEADER: Node Name + Node Mode Icon */}
```
surchs: this comment (and the others like it) are a good signal that the code under the comment should be a component. I'd say that if you need to do special JSX escape stuff so you can add comments to structure the JSX bit, then the JSX bit is too long and complex (it is) and we should factor out the components. Seems to me the `ResultCard` component is essentially going to be a container component that keeps state and assembles a bunch of UI components.

src/components/ResultContainer.tsx
```diff
@@ -206,24 +220,25 @@ function ResultContainer({
...
                 )
-              : parsePipelinesInfoToString(res.available_pipelines),
+              : parsePipelinesInfoToString(datasetPipelines),
+            accessLink,
           ].join('\t')
         );
       } else {
         // @ts-expect-error: typescript doesn't know that subject_data is an array when records_protected is false.
-        res.subject_data.forEach((subject) => {
+        subResp.subject_data.forEach((subject) => {
```
surchs, line 239: Just noticed this: the `Subject` type definition is missing a couple of keys that you use here, i.e. `session_typ`, `num_matching_phenotypic_sessions` and `num_matching_imaging_sessions`. The reason this isn't caught by ts is because you have this `@ts-expect-error` here. That's not great - could you please open an issue to follow up and check if there is another way to handle this. We would like to have the type safety here, i.e. get errors if we forget things in the interfaces

src/components/ResultContainer.tsx
```diff
@@ -183,16 +183,30 @@ function ResultContainer({
...
       'DatasetImagingModalities',
       'DatasetPipelines',
+      'AccessLink',
     ].join('\t');
     tsvRows.push(headers);
 
-    subjectsData.responses.forEach((res) => {
-      if (res.records_protected) {
+    subjectsResponse.responses.forEach((subResp) => {
+      const datasetMetadata = datasetsResponse?.responses.find(
```
surchs, line 192: this isn't new, but since I'm looking at it now: we're mutating `tsvRows` here in the `forEach`. That caused us some grief recently over in the annotation tool. Let's at least add a TODO to address this in a future refactor

src/components/ResultContainer.tsx
```diff
@@ -183,16 +183,30 @@ function ResultContainer({
...
+
+      // Fallback values if merge fails (should not happen if UUIDs match)
+      const datasetName = datasetMetadata?.dataset_name ?? '';
+      const repositoryUrl = datasetMetadata?.repository_url ?? '';
+      const accessLink = datasetMetadata?.access_link ?? '';
+      const numMatchingSubjects = datasetMetadata?.num_matching_subjects ?? 0;
+      const isProtected = datasetMetadata?.records_protected ?? false;
+      const datasetImageModals = datasetMetadata?.image_modals ?? [];
+      const datasetPipelines = datasetMetadata?.available_pipelines ?? {};
+
```
surchs, line 206: I think this can be simpler, i.e. we don't need to check each attribute. if `datasetMetadata` is falsy, then we just destruct the entire object of fallback defaults. That should also read more easily

surchs (changes requested):
Thanks @rmanaem for the PR and the very nice looking UI!

Two main points for the request changes:
1. I think your retry-await construct in the e2e test is incorrect. Please double-check the cypress docs on sync vs async and then separately see if we actually need to reinvent the retry
2. the ResultCard component is getting pretty unwieldy, please factor out the components you already distinguish with comments and inline functions so we have a fighting chance when we get back to it next time.

cypress/e2e/UpdateExamples.cy.ts
surchs: edit: I should have commented on the next `check()` because the next check is synchronous. This check here is in fact inside the async `.then()`, but that just means some of the code is queued when the rest of the code is run during parsing. Not completely sure how this plays out in practice but I strongly suspect it's wrong

README.md
rmanaem: Ah yes I forgot we added the access link as well. Will include it.

cypress/component/ResultCard.cy.tsx
rmanaem: I'd say this is more explicit and therefore clearer.

cypress/e2e/UpdateExamples.cy.ts
rmanaem: I believe we can't use the retry ability because cy.task is an action which executes once and yields a result.

> Seems to me that maybe all it takes is to add an assertion like .should('contain', 'something')

Just so I understand, do u mean should simply check if the download dir contains the file?

src/App.tsx
rmanaem: Yea ircc this was to address the Form e2e test flakiness.

cypress/e2e/UpdateExamples.cy.ts
rmanaem: I've gone ahead and used cy.clock to "mock" the date/time and therefore the timestamp string:

```TS
cy.clock(Date.UTC(2025, 0, 1), ['Date']);

    cy.readFile(
      'cypress/downloads/neurobagel-query-results_20250101000000.tsv'
    ).should('exist');

    cy.get('[data-cy="download-radio-1"]').click();

    cy.readFile(
      'cypress/downloads/neurobagel-query-results-with-URIs_20250101000000.tsv'
    ).should('exist');
```

src/components/ResultContainer.tsx
rmanaem: In the annotation tool? I don't recall.

cypress/e2e/UpdateExamples.cy.ts
surchs: I mean that if you look at both

> Assertions are automatically retried until they pass or time out.

https://docs.cypress.io/api/commands/should

and 

> If you are using commands, but need to retry the entire chain, consider rewriting the commands into a [.should(callbackFn)](https://docs.cypress.io/api/commands/should#Function).

https://docs.cypress.io/app/core-concepts/retry-ability#Use-should-with-a-callback

it seems to me that you should be able to just do 

```js
cy.task('getLatestFile', pattern).should('something);
```

to get the retry for free from cypress.

src/components/ResultContainer.tsx
rmanaem: Ah good catch!
I'll update the type as well.

cypress/e2e/UpdateExamples.cy.ts
surchs: > I've gone ahead and used cy.clock to "mock" the date/time and therefore the timestamp string:

OK, that should also work. But since we're already dealing with this problem, please also check the sync/async part of cypress to make sure we know how to handle this scenario correctly next time. It is pretty tricky, and now we know as much about it as we will for a while

src/components/ResultContainer.tsx
rmanaem: Here is the issue: https://github.com/neurobagel/query-tool/issues/702

cypress/e2e/UpdateExamples.cy.ts
rmanaem: Yea thanks for mentioning that, I wasn't aware of that. I've noted it down to read through it.

src/components/ResultCard/DatasetInfoColumn.tsx
```diff
@@ -0,0 +1,104 @@
...
+  checked,
+  onCheckboxChange,
+  datasetName,
+  authors,
+  homepage,
+  repositoryUrl,
+  accessType,
+}: DatasetInfoColumnProps) {
+  return (
+    <div className="col-span-5 flex gap-2">
```
surchs: these placement styles should live with the parent because that's also who is responsible for the (external) layout. The component can own relative placement of its subparts and coloring and such things.

src/components/ResultContainer.tsx
surchs: I think you might be right that I made that up or am confusing with something else in store refactor. Can't recall exactly

src/components/ResultContainer.tsx
```diff
@@ -183,16 +183,33 @@ function ResultContainer({
...
+        dataset_name: datasetName = '',
+        repository_url: repositoryUrl = '',
+        access_link: accessLink = '',
+        num_matching_subjects: numMatchingSubjects = 0,
+        records_protected: isProtected = false,
+        image_modals: datasetImageModals = [],
+        available_pipelines: datasetPipelines = {},
+      } = datasetMetadata || {};
+
+      if (isProtected) {
```
surchs, line 207: neat

surchs (approved):
Thanks @rmanaem, this looks good. I'm approving, but please take a look at the layouting of `ResultCard`. I think we should make the grid positioning clearly the responsibility of the `ResultCard` component and not of each child. 

🧑‍🍳

PR #714, opened 2026-01-20
[FIX] Correct the position of `ErrorBoundary` in the app to catch render errors

### LINKED ISSUE #645: Investigate why the error boundary did not catch error
### Root cause
Errors are not being caught by our ErrorBoundary because it is wrapper outside the Router. React Router has its own ErrorBoudnary [docs](https://reactrouter.com/how-to/error-boundary):

> To avoid rendering an empty page to users, route modules will automatically catch errors in your code and render the closest ErrorBoundary.

which catches the errors first and shows the generic "Unexpected Application Error" before they could reach our ErrorBoundary.

### To Address
- [x] Add a regression test
- [x] Move the Error Boundary inside the router in main.tsx

The current order of wrappers in `main` where 

### Expected behavior

We just encountered an app-breaking bug in #642. We would expect this to be caught by the error boundary.

### Current behavior

The app just crashed out completely

### Error message

_No response_

### Environment

- OS:
- Python/Node version:
- Docker version:
- Docker Compose version:

### How to reproduce

_No response_

### Anything else?

_No response_

rmanaem:
- Closes #645 

- Added a regression test for error boundary component
- Wrapped App component in ErrorBoundary

## Checklist

- [x] Tests pass
- [x] Checks pass

For new features:
- [ ] Tests have been added

For bug fixes:
- [x] There is at least one test that would fail under the original bug conditions.

## Summary by Sourcery

Move the error boundary inside the router so application-level errors are correctly caught and display the fallback UI, and add coverage to prevent regressions.

Bug Fixes:
- Ensure errors occurring within routed application views are caught by the error boundary and render the fallback screen instead of crashing.

Enhancements:
- Expose the error boundary container via a data attribute to support reliable E2E selection in tests.

Tests:
- Add an end-to-end regression test using a mocked response with missing fields to verify the error boundary is triggered and its fallback UI is shown when the app crashes.


cypress/e2e/ErrorBoundaryRegression.cy.ts
```diff
@@ -0,0 +1,24 @@
+import { missingFieldsResponse } from '../fixtures/mocked-responses';
+
+describe('Error Boundary Regression Test', () => {
+  it('Should catch application crashes with the Custom Error Boundary', () => {
```
surchs, line 4: nice

cypress/fixtures/mocked-responses.ts
```diff
@@ -511,3 +511,29 @@ export const responseWithIncompleteModality = {
...
+// Response with missing fields (for error boundary testing)
+export const missingFieldsResponse = {
+  nodes_response_status: 'success',
+  errors: [],
+  responses: [
+    {
+      node_name: 'Test Node',
+      dataset_uuid: '123',
+      dataset_name: 'Crash Dataset',
+      // authors is missing!
```
surchs, line 523: :tada:

surchs (approved):
Great stuff, thanks @rmanaem 

:cook:

PR #841, opened 2026-08-20
[ENH] Enable selecting multiple options for AND queries

### LINKED ISSUE #832: Enable selecting multiple options for AND queries
Implement multi-select for: diagnosis, assessment, imaging modality, pipeline name excluding sex b/c each subject can only have 1 sex

Pipeline version: we would need to group pipeline versions by pipeline name somehow, either with a simple prefix (easiest) or as dropdown option sections

Update POST body for queries

To implement:

- [x] Update `QueryParams` and `SubjectsRequestBody` types in `src/utils/types.ts` to support multi-select field structures.
- [x] Enable `multiple` selection prop for categorical fields in `src/components/QueryForm.tsx`.
- [x] Combine pipeline name and version fields into a single "Pipeline name and version" collapsible tree-select dropdown.
- [x] Group pipeline version options under collapsible pipeline group accordions, collapsed by default.
- [x] Format pipeline group header selection chips as `<Pipeline Name> any version`.
- [x] Auto-deselect pipeline versions when their parent pipeline group header is unchecked or removed.
- [x] Update version fetching logic in `src/App.tsx` to retrieve versions for all selected/available pipelines.
- [x] Update `constructDatasetsRequestBody` in `src/App.tsx` to structure array query parameters (`diagnosis`, `assessment`, `image_modal`, `pipeline`).
- [x] Update form equality checks and utility helpers in `src/utils/utils.ts`.
- [x] Extract `CollapsiblePipelineGroup` into a standalone component with dedicated component tests (`CollapsiblePipelineGroup.cy.tsx`).
- [x] Update Cypress component (`QueryForm.cy.tsx`, `CategoricalField.cy.tsx`) and E2E test suites (`Form.cy.ts`, `APIRequests.cy.ts`) for multi-select fields and request payloads.

rmanaem:
- Closes #832

  - Enabled multi-selection across diagnosis, assessment, imagingModality, and pipeline fields.
  - Combined separate Pipeline Name and Pipeline Version inputs into a single "Pipeline name and version" collapsible tree-select dropdown, collapsed by default.
  - Formatted pipeline group header selections to display as `<Pipeline Name> any version` chip pills (e.g., `fmriprep any version`).
  - Configured header unchecking to automatically deselect child pipeline versions.
  - Updated `constructDatasetsRequestBody` in `src/App.tsx` and API payload types to format multi-select fields (diagnosis, assessment, image_modal, pipeline) as array payloads.
  - Refactored form logic by extracting `CollapsiblePipelineGroup` into a standalone component with explicit props (`groupKey`, `groupLabel`, `children`).
  - Centralized pure pipeline option transformers and numeric validation helpers in `src/utils/utils.ts`.
  - Added dedicated Cypress component tests for `CollapsiblePipelineGroup` (`CollapsiblePipelineGroup.cy.tsx`) and updated existing component and E2E test suites (`Form.cy.ts`, `APIRequests.cy.ts`, `QueryForm.cy.tsx`).

## Checklist

- [x] Tests pass
- [x] Checks pass

For new features:
- [ ] Tests have been added

For bug fixes:

## Summary by Sourcery

Enhance query filtering with multi-select categorical fields and a unified grouped pipeline name/version selector.

New Features:
- Combine pipeline names and versions into a grouped, multi-select query field with support for selecting any version or specific versions.
- Support multiple selections for diagnosis, assessment tools, and imaging modalities.

Bug Fixes:
- Preserve valid pipeline version selections when pipeline selections change and clear incompatible versions.

Enhancements:
- Update query construction to submit categorical filters as arrays and pipelines as name/version pairs.
- Load pipeline versions for all available pipelines and provide collapsible pipeline groups with checkbox controls.

Tests:
- Add component and end-to-end coverage for multi-select categorical fields, grouped pipeline selection, query payloads, clearing selections, and pipeline version loading.

## Summary by Sourcery

Enable multi-value query filtering with a unified pipeline name and version selector.

New Features:
- Enable multi-selection for diagnosis, assessment, and imaging modality query filters.
- Provide a unified grouped pipeline name and version selector supporting any-version and specific-version choices.

Bug Fixes:
- Clear incompatible pipeline version selections when their pipeline is deselected or changed.

Enhancements:
- Submit multi-select categorical filters as arrays and pipeline filters as name/version query objects.
- Load pipeline versions across available pipelines and present them in collapsible grouped selections.

Tests:
- Add component and end-to-end coverage for multi-select fields, grouped pipeline selection, query payloads, and selection clearing.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -307,19 +307,23 @@ function App() {
       }
     }
 
-    const selectedPipelineNames = normalizeFieldInputOptions(pipelineName);
+    const pipelineURIs = Object.keys(pipelines);
 
-    selectedPipelineNames.forEach((p) => {
-      if (pipelines[p.id] == null || pipelines[p.id].length === 0) {
-        getPipelineVersions(p).then((pipelineVersionsResponse) => {
+    pipelineURIs.forEach((pId) => {
+      if (pipelines[pId] == null || pipelines[pId].length === 0) {
+        const pOption: FieldInputOption = {
+          id: pId,
+          label: pId.startsWith('np:') ? pId.slice(3) : pId,
+        };
+        getPipelineVersions(pOption).then((pipelineVersionsResponse) => {
           setPipelines((prevPipelines) => ({
             ...prevPipelines,
-            [p.id]: pipelineVersionsResponse,
+            [pId]: pipelineVersionsResponse,
           }));
         });
       }
     });
-  }, [pipelines, pipelineName]);
+  }, [pipelines]);
 
   useEffect(() => {
     if (availableNodes.length > 1) {
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -36,6 +36,7 @@ import NodeAdmonition from './components/NodeAdmonition';
 import './App.css';
 import logo from './assets/logo.png';
 import areFormStatesEqual, {
+  normalizeFieldInputOptions,
   parseNumericValue,
   sendDatasetsQuery,
   sendSubjectsQuery,
@@ -306,11 +307,7 @@ function App() {
       }
     }
 
-    const selectedPipelineNames: FieldInputOption[] = Array.isArray(pipelineName)
-      ? pipelineName
-      : pipelineName
-        ? [pipelineName]
-        : [];
+    const selectedPipelineNames = normalizeFieldInputOptions(pipelineName);
 
     selectedPipelineNames.forEach((p) => {
       if (pipelines[p.id] == null || pipelines[p.id].length === 0) {
@@ -405,16 +402,10 @@ function App() {
         break;
       case 'Pipeline name': {
         setPipelineName(value);
-        const newSelectedPipelines: FieldInputOption[] = Array.isArray(value)
-          ? value
-          : value
-            ? [value]
-            : [];
+        const newSelectedPipelines = normalizeFieldInputOptions(value);
         const newPipelineIds = new Set(newSelectedPipelines.map((p) => p.id));
         if (pipelineVersion) {
-          const currentVersions: FieldInputOption[] = Array.isArray(pipelineVersion)
-            ? pipelineVersion
-            : [pipelineVersion];
+          const currentVersions = normalizeFieldInputOptions(pipelineVersion);
           const validVersions = currentVersions.filter((v) => {
             const pId = v.id.includes('::') ? v.id.split('::')[0] : '';
             return newPipelineIds.has(pId);
@@ -470,11 +461,9 @@ function App() {
     if (maxAgeNumber !== null) requestBody.max_age = maxAgeNumber;
     if (sex && !Array.isArray(sex)) requestBody.sex = sex.id;
 
-    if (diagnosis) {
-      const diagArray = Array.isArray(diagnosis) ? diagnosis : [diagnosis];
-      if (diagArray.length > 0) {
-        requestBody.diagnosis = diagArray.map((d) => d.id);
-      }
+    const diagArray = normalizeFieldInputOptions(diagnosis);
+    if (diagArray.length > 0) {
+      requestBody.diagnosis = diagArray.map((d) => d.id);
     }
 
     const minNumImagingSessionsNumber = parseNumericValue(minNumImagingSessions);
@@ -485,32 +474,20 @@ function App() {
     if (minNumPhenotypicSessionsNumber !== null)
       requestBody.min_num_phenotypic_sessions = minNumPhenotypicSessionsNumber;
 
-    if (assessmentTool) {
-      const assessArray = Array.isArray(assessmentTool) ? assessmentTool : [assessmentTool];
-      if (assessArray.length > 0) {
-        requestBody.assessment = assessArray.map((a) => a.id);
-      }
+    const assessArray = normalizeFieldInputOptions(assessmentTool);
+    if (assessArray.length > 0) {
+      requestBody.assessment = assessArray.map((a) => a.id);
     }
 
-    if (imagingModality) {
-      const modalArray = Array.isArray(imagingModality) ? imagingModality : [imagingModality];
-      if (modalArray.length > 0) {
-        requestBody.image_modal = modalArray.map((m) => m.id);
-      }
+    const modalArray = normalizeFieldInputOptions(imagingModality);
+    if (modalArray.length > 0) {
+      requestBody.image_modal = modalArray.map((m) => m.id);
     }
 
-    const selectedPipelines: FieldInputOption[] = Array.isArray(pipelineName)
-      ? pipelineName
-      : pipelineName
-        ? [pipelineName]
-        : [];
+    const selectedPipelines = normalizeFieldInputOptions(pipelineName);
 
     if (selectedPipelines.length > 0) {
-      const selectedVersions: FieldInputOption[] = Array.isArray(pipelineVersion)
-        ? pipelineVersion
-        : pipelineVersion
-          ? [pipelineVersion]
-          : [];
+      const selectedVersions = normalizeFieldInputOptions(pipelineVersion);
 
       requestBody.pipeline = selectedPipelines.flatMap((p) => {
         const pVersions = selectedVersions.filter(
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -305,21 +305,23 @@ function App() {
         return [];
       }
     }
-    // Get pipeline versions if
-    // 1. A pipeline has been selected (this implementation only works for single value for pipeline name field)
-    // 2. This is the first time its being selected (i.e., we haven't retrieved pipeline versions before)
-    if (
-      pipelineName !== null &&
-      !Array.isArray(pipelineName) &&
-      pipelines[pipelineName.id].length === 0
-    ) {
-      getPipelineVersions(pipelineName).then((pipelineVersionsRespnse) => {
-        setPipelines((prevPipelines) => ({
-          ...prevPipelines,
-          [pipelineName.id]: pipelineVersionsRespnse,
-        }));
-      });
-    }
+
+    const selectedPipelineNames: FieldInputOption[] = Array.isArray(pipelineName)
+      ? pipelineName
+      : pipelineName
+        ? [pipelineName]
+        : [];
+
+    selectedPipelineNames.forEach((p) => {
+      if (pipelines[p.id] == null || pipelines[p.id].length === 0) {
+        getPipelineVersions(p).then((pipelineVersionsResponse) => {
+          setPipelines((prevPipelines) => ({
+            ...prevPipelines,
+            [p.id]: pipelineVersionsResponse,
+          }));
+        });
+      }
+    });
   }, [pipelines, pipelineName]);
 
   useEffect(() => {
@@ -401,9 +403,26 @@ function App() {
       case 'Pipeline version':
         setPipelineVersion(value);
         break;
-      case 'Pipeline name':
+      case 'Pipeline name': {
         setPipelineName(value);
+        const newSelectedPipelines: FieldInputOption[] = Array.isArray(value)
+          ? value
+          : value
+            ? [value]
+            : [];
+        const newPipelineIds = new Set(newSelectedPipelines.map((p) => p.id));
+        if (pipelineVersion) {
+          const currentVersions: FieldInputOption[] = Array.isArray(pipelineVersion)
+            ? pipelineVersion
+            : [pipelineVersion];
+          const validVersions = currentVersions.filter((v) => {
+            const pId = v.id.includes('::') ? v.id.split('::')[0] : '';
+            return newPipelineIds.has(pId);
+          });
+          setPipelineVersion(validVersions.length > 0 ? validVersions : null);
+        }
         break;
+      }
       default:
         break;
     }
@@ -450,7 +469,14 @@ function App() {
     const maxAgeNumber = parseNumericValue(maxAge);
     if (maxAgeNumber !== null) requestBody.max_age = maxAgeNumber;
     if (sex && !Array.isArray(sex)) requestBody.sex = sex.id;
-    if (diagnosis && !Array.isArray(diagnosis)) requestBody.diagnosis = diagnosis.id;
+
+    if (diagnosis) {
+      const diagArray = Array.isArray(diagnosis) ? diagnosis : [diagnosis];
+      if (diagArray.length > 0) {
+        requestBody.diagnosis = diagArray.map((d) => d.id);
+      }
+    }
+
     const minNumImagingSessionsNumber = parseNumericValue(minNumImagingSessions);
     if (minNumImagingSessionsNumber !== null)
       requestBody.min_num_imaging_sessions = minNumImagingSessionsNumber;
@@ -458,13 +484,48 @@ function App() {
     const minNumPhenotypicSessionsNumber = parseNumericValue(minNumPhenotypicSessions);
     if (minNumPhenotypicSessionsNumber !== null)
       requestBody.min_num_phenotypic_sessions = minNumPhenotypicSessionsNumber;
-    if (assessmentTool && !Array.isArray(assessmentTool))
-      requestBody.assessment = assessmentTool.id;
-    if (imagingModality && !Array.isArray(imagingModality))
-      requestBody.image_modal = imagingModality.id;
-    if (pipelineName && !Array.isArray(pipelineName)) requestBody.pipeline_name = pipelineName.id;
-    if (pipelineVersion && !Array.isArray(pipelineVersion) && pipelineName)
-      requestBody.pipeline_version = pipelineVersion.id;
+
+    if (assessmentTool) {
+      const assessArray = Array.isArray(assessmentTool) ? assessmentTool : [assessmentTool];
+      if (assessArray.length > 0) {
+        requestBody.assessment = assessArray.map((a) => a.id);
+      }
+    }
+
+    if (imagingModality) {
+      const modalArray = Array.isArray(imagingModality) ? imagingModality : [imagingModality];
+      if (modalArray.length > 0) {
+        requestBody.image_modal = modalArray.map((m) => m.id);
+      }
+    }
+
+    const selectedPipelines: FieldInputOption[] = Array.isArray(pipelineName)
+      ? pipelineName
+      : pipelineName
+        ? [pipelineName]
+        : [];
+
+    if (selectedPipelines.length > 0) {
+      const selectedVersions: FieldInputOption[] = Array.isArray(pipelineVersion)
+        ? pipelineVersion
+        : pipelineVersion
+          ? [pipelineVersion]
+          : [];
+
+      requestBody.pipeline = selectedPipelines.flatMap((p) => {
+        const pVersions = selectedVersions.filter(
+          (v) =>
+            v.id.startsWith(`${p.id}::`) || (selectedPipelines.length === 1 && !v.id.includes('::'))
+        );
+        if (pVersions.length > 0) {
+          return pVersions.map((v) => {
+            const versionStr = v.id.includes('::') ? v.id.split('::')[1] : v.id;
+            return { name: p.id, version: versionStr };
+          });
+        }
+        return [{ name: p.id }];
+      });
+    }
 
     return requestBody;
   }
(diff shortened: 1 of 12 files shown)


src/App.tsx
```diff
@@ -450,21 +464,46 @@ function App() {
...
-    if (pipelineName && !Array.isArray(pipelineName)) requestBody.pipeline_name = pipelineName.id;
-    if (pipelineVersion && !Array.isArray(pipelineVersion) && pipelineName)
-      requestBody.pipeline_version = pipelineVersion.id;
+
+    const assessArray = normalizeFieldInputOptions(assessmentTool);
+    if (assessArray.length > 0) {
+      requestBody.assessment = assessArray.map((a) => a.id);
+    }
+
+    const modalArray = normalizeFieldInputOptions(imagingModality);
```
surchs: Variable name makes no sense any more. Accidentally refactor by bot?

src/App.tsx
```diff
@@ -450,21 +464,46 @@ function App() {
...
       requestBody.min_num_phenotypic_sessions = minNumPhenotypicSessionsNumber;
-    if (assessmentTool && !Array.isArray(assessmentTool))
-      requestBody.assessment = assessmentTool.id;
-    if (imagingModality && !Array.isArray(imagingModality))
-      requestBody.image_modal = imagingModality.id;
-    if (pipelineName && !Array.isArray(pipelineName)) requestBody.pipeline_name = pipelineName.id;
-    if (pipelineVersion && !Array.isArray(pipelineVersion) && pipelineName)
-      requestBody.pipeline_version = pipelineVersion.id;
+
+    const assessArray = normalizeFieldInputOptions(assessmentTool);
```
surchs: Here too: variable name makes no sense any more

src/App.tsx
```diff
@@ -450,21 +464,46 @@ function App() {
...
+
+    const selectedPipelines = normalizeFieldInputOptions(pipelineName);
+
+    if (selectedPipelines.length > 0) {
+      const selectedVersions = normalizeFieldInputOptions(pipelineVersion);
+
+      requestBody.pipeline = selectedPipelines.flatMap((p) => {
+        const pVersions = selectedVersions.filter((v) => v.id.startsWith(`${p.id}::`));
+        if (pVersions.length > 0) {
+          return pVersions.map((v) => ({
```
surchs, line 505: I'd say this is a pretty good indicator that we shouldn't have done this "hierarchy is encoded as `::` string concatenation".

src/App.tsx
surchs: i.e. this is imagingmodality, not a modal

src/components/QueryForm.tsx
```diff
@@ -99,6 +93,75 @@ function QueryForm({
...
+
+    updateCategoricalQueryParams(
+      'Pipeline name',
+      updatedPipelines.length > 0 ? updatedPipelines : null
+    );
+    updateCategoricalQueryParams(
+      'Pipeline version',
+      updatedVersions.length > 0 ? updatedVersions : null
+    );
+  };
```
surchs: afaict these are all concerns that the select-component could handle for me. I.e. You are a pipeline / hierarchical select component (stateless / display only). I give you a list of pipelines and versions to show. When the user clicks on any of them, you tell me about it by giving me the pipeline and version they clicked (e.g. version is empty if they clicked on the "any version" option).

surchs (changes requested):
Hey @rmanaem , thanks for working on this.

I ran out of time during my review, so I will submit before commenting everywhere. High level I think this is trying to force the new requirements into the old implementation. I.e. we now have a CategoricalField component that handles 3 different scenarios (single-select, i.e. Sex - hierarchical-select, i.e. Pipelines - and multi-select, i.e. the rest). I'd say they should all be separate components (with shared code factored out of course / reuse as possible).

afaict this would allow you to get rid of some of the self-imposed limitations here that made you encode pipeline hierarchy as a flat list with `::` separators - which you then later have to decode again to handle the request building - and some of the other helper scripts you wrote. And we could also get rid of the somewhat confusing dropdown behaviour for Sex right now with a checkbox.

So could you please refactor into separate components for the (currently) 3 cases to allow for the new capabilities of multi-selection and , and make sure to reuse as possible

src/components/QueryForm.tsx
```diff
@@ -99,18 +91,69 @@ function QueryForm({
...
+    const updatedPipelines = Array.from(updatedPipelinesMap.values());
+    updateCategoricalQueryParams(
+      'Pipeline name',
+      updatedPipelines.length > 0 ? updatedPipelines : null
+    );
+    updateCategoricalQueryParams(
+      'Pipeline version',
+      updatedVersions.length > 0 ? updatedVersions : null
+    );
+  };
```
surchs: **review**: check why handler cannot live in pipeline field component

src/App.tsx
```diff
@@ -55,6 +57,9 @@ function App() {
     { NodeName: 'All', ApiURL: 'allNodes' },
   ]);
   const [pipelines, setPipelines] = useState<Pipelines>({});
+  // Track pipeline IDs for which version fetching has already been initiated to prevent
+  // duplicate in-flight requests or infinite fetch loops when a pipeline has no versions.
+  const fetchedPipelineURIs = useRef<Set<string>>(new Set());
```
surchs: **Review ** can we reuse exist types here?

src/components/CollapsiblePipelineGroup.tsx
```diff
@@ -0,0 +1,74 @@
...
+              my: 0,
+              alignItems: 'center',
+              '&.Mui-expanded': { my: 0 },
+            },
+          }}
+        >
+          <Checkbox
+            data-cy={`pipeline-group-${pipelineId}-checkbox`}
+            size="small"
+            checked={isPipelineChecked}
```
surchs: I suggest also handling "indeterminate" as a checkbox state when specific versions are selected in a pipeline (i.e. not the entire pipeline or "any version", but only specific versions of the pipeline): https://mui.com/material-ui/react-checkbox/#indeterminate

Otherwise the pipeline checkbox appears unchecked when collapsed - even if individual pipeline versions are selected

src/components/PipelineField.tsx
```diff
@@ -0,0 +1,118 @@
...
+    }
+    return versions.map((v) => ({
+      pipelineId: pId,
+      pipelineLabel: pLabel,
+      version: v,
+    }));
+  });
+}
+
+function PipelineField({ pipelines, value, onFieldChange, disabled = false }: PipelineFieldProps) {
```
surchs: Mind adding some docstring to these components (here and for the other ones) - it's not immediately obvious to me from the component / file name what the component does

src/components/CollapsiblePipelineGroup.tsx
surchs: I may have misread this component - my comment refers to the top-level checkbox, i.e. the "fmriprep" checkbox, not the "fmriprep v27.0.1" checkbox

src/components/PipelineField.tsx
surchs: PipelineSelectField? Or more generically "HierarchicalSelectField"?

src/components/PipelineField.tsx
```diff
@@ -0,0 +1,118 @@
...
+export interface PipelineOption {
+  pipelineId: string;
+  pipelineLabel: string;
+  version?: string;
+}
+
+export interface PipelineFieldProps {
+  pipelines: Pipelines;
+  value: PipelineOption[];
+  onFieldChange: (selectedPipelines: PipelineOption[]) => void;
```
surchs: I think this component could / should be more generic than "pipeline". But if you want to keep it specific to pipeline, then this should also be called `selectablePipelines` and 'selectedPipelines' or something like that. `pipelines` and `value` is confusing

src/components/PipelineField.tsx
```diff
@@ -0,0 +1,118 @@
...
+    return versions.map((v) => ({
+      pipelineId: pId,
+      pipelineLabel: pLabel,
+      version: v,
+    }));
+  });
+}
+
+function PipelineField({ pipelines, value, onFieldChange, disabled = false }: PipelineFieldProps) {
+  const options = buildPipelineOptions(pipelines);
```
surchs: similarly to above: I suggest keeping this component more generic as a "hierachical multi select" component. This builder function is coupling the component tightly to our internal representations of "pipeline options". This should be the responsibility of the parent who should figure out how to feed this (presentational) component with the correct input data.

src/components/PipelineField.tsx
```diff
@@ -0,0 +1,118 @@
...
+
+  const handleTogglePipeline = (pId: string, pLabel: string) => {
+    const isHeaderChecked = value.some((opt) => opt.pipelineId === pId && !opt.version);
```
surchs: iiuc this checks if a top-level item is selected in a hierarchical tree. How about `isTopLevelOptionChecked` instead? Header to me sounds like a UI element but this refers to the hierarchy of options

src/components/PipelineField.tsx
```diff
@@ -0,0 +1,118 @@
...
+      value={value}
+      isOptionEqualToValue={(option, val) =>
+        option.pipelineId === val.pipelineId && option.version === val.version
+      }
+      getOptionLabel={(option) =>
+        option.version
+          ? `${option.pipelineLabel} ${option.version}`
+          : `${option.pipelineLabel} any version`
+      }
+      groupBy={(option) => option.pipelineLabel}
```
surchs: why not just make this an actual field in the options passed from the parent? i.e. option.label? Then this component could be more generic, there wouldn't be important logic sitting in the render section, and we could more easily change what the label is in one location.

src/components/PipelineField.tsx
```diff
@@ -0,0 +1,118 @@
...
+      renderGroup={({ key, group, children }) => {
+        const pId =
+          Object.keys(pipelines).find(
+            (id) => (id.startsWith('np:') ? id.slice(3) : id) === group
+          ) ?? `np:${group}`;
+        const isPipelineChecked = value.some((opt) => opt.pipelineId === pId && !opt.version);
+
```
surchs: so iiuc you have to do this somewhat hacky reconstruction of the hierarchy because you group by the pipelineLabel rather than by the pipelineId. If you instead grouped by pipelineID (or a generalizable attribute like "parentID"), you would just have to get the group label here. But that should be easy to do since it's already in each option (currently as option.pipelineLabel). So something like `options.find((o) => o.pipelineId === group)?.pipelineLabel ?? "don't know";` should do

src/components/PipelineField.tsx
```diff
@@ -0,0 +1,118 @@
...
+      renderInput={(params) => (
+        <TextField {...params} label="Pipeline name and version" placeholder="Select an option" />
+      )}
+      renderOption={(props, option, { selected }) => (
+        <li {...props} key={`${option.pipelineId}-${option.version}`}>
+          <Checkbox size="small" sx={{ mr: 1 }} checked={selected} />
+          {option.version
+            ? `${option.pipelineLabel} ${option.version}`
+            : `${option.pipelineLabel} any version`}
+        </li>
```
surchs: duplicated here again, see my above comment about just pre-computing this (at the parent) in the input structure

src/components/QueryForm.tsx
```diff
@@ -99,18 +91,69 @@ function QueryForm({
...
+      }));
+    }
+    return [
+      {
+        pipelineId: p.id,
+        pipelineLabel: p.label,
+      },
+    ];
+  });
+
```
surchs: I think this right here is the problem with how state is managed. A pipeline has a name and a version (and some other stuff). Because `selectedPipelines` and `selectedVersions` are separate states, there is all of this duplicated code in a bunch of places to stitch them together and then split them apart again.

src/components/QueryForm.tsx
```diff
@@ -99,18 +91,69 @@ function QueryForm({
...
+  const handlePipelineFieldChange = (selectedOptions: PipelineOption[]) => {
+    if (selectedOptions.length === 0) {
+      updateCategoricalQueryParams('Pipeline name', null);
+      updateCategoricalQueryParams('Pipeline version', null);
+      return;
```
surchs: same here - the fact that name and version are different query parameters should not leak through the whole app / drive the state. This can be split once when we actually build the query POST body

src/App.tsx
```diff
@@ -55,6 +57,9 @@ function App() {
     { NodeName: 'All', ApiURL: 'allNodes' },
   ]);
   const [pipelines, setPipelines] = useState<Pipelines>({});
```
surchs: I think what we would need here is state for `selectedPipelines`. And then that's the state where we keep track of the exact pipelines+ their versions that are currently selected!

surchs: And the state below 

```
  const [pipelineVersion, setPipelineVersion] = useState<FieldInput>(null);
  const [pipelineName, setPipelineName] = useState<FieldInput>(null);
```
Should be removed

surchs (changes requested):
Thanks @rmanaem for addressing the component comments. 

Sorry for not having read through the entire diff last time - but I am requesting changes again. There is a good amount of duplicated logic in this PR (and in the app already) that I believe is all trickling down from the (so far maybe somewhat reasonable) idea to handle pipeline name and pipeline version as separate states - essentially making the query parameters drive the app state. 

Some examples:

- https://github.com/neurobagel/query-tool/pull/841/changes#diff-e6ef9513f0659e25bed83675da45b9877cec212e8f28cc95790ff9055351f35cR94-R112
- https://github.com/neurobagel/query-tool/pull/841/changes#diff-e6ef9513f0659e25bed83675da45b9877cec212e8f28cc95790ff9055351f35cR114-R133
- https://github.com/neurobagel/query-tool/pull/841/changes#diff-26ad4b834941d9b19ebf9db8082bd202aaf72ea0ddea85f5a8a0cb3c729cc6f2R417-R424
- https://github.com/neurobagel/query-tool/pull/841/changes#diff-26ad4b834941d9b19ebf9db8082bd202aaf72ea0ddea85f5a8a0cb3c729cc6f2R497-R512
- https://github.com/neurobagel/query-tool/pull/841/changes#diff-5ff75eaa3b0fee350fc9d5d1840bc4b311acc99b46b3d6222b02f6b0c6c2be67R81-R85

I haven't completely untangled this in my mind, but I think it's pretty clear that if we handled `selectedPipelines` as top level state (i.e. in App.tsx) once, and then pass pipeline and version together through the app, and only split it once when we build the query parameters - this should be a good deal cleaner

Happy to think through this together. But I think we have to change this now (because a lot of duplication is being added here, not just old stuff).

src/components/PipelineField.tsx
rmanaem: I prefer `PipelineSelectField` for now.

src/utils/utils.ts
```diff
@@ -11,12 +15,109 @@ import {
...
+
+export function getPipelineLabel(pId: string): string {
+  return pId.startsWith('np:') ? pId.slice(3) : pId;
+}
+
+export function buildPipelineOptions(pipelines: Pipelines): HierarchicalOption[] {
+  return Object.keys(pipelines).flatMap((pId): HierarchicalOption[] => {
+    const pLabel = getPipelineLabel(pId);
+    const versions = pipelines[pId] ?? [];
+    if (versions.length === 0) {
```
surchs: I don't understand how you can ever get to `pipelines[pId]` being nullish and/or versions.length === 0

surchs: More importantly, unless it is completely obvious and I have missed it, this is probably not a good way to encode "top-level-ness" given how hard it is to reason about!

src/components/HierarchicalSelectField.tsx
```diff
@@ -0,0 +1,133 @@
...
+      multiple
+      disableCloseOnSelect
+      data-cy={dataCy}
+      options={options}
+      value={value}
+      isOptionEqualToValue={(option, val) => option.id === val.id}
+      getOptionLabel={(option) => option.label}
+      groupBy={(option) => option.parentId}
+      renderGroup={({ key, group, children }) => {
+        const parentLabel = parentLabelMap[group] ?? group;
```
surchs: ## review
Map is made by disassembling the pipeline Options again to extract the labels

src/components/HierarchicalSelectField.tsx
```diff
@@ -0,0 +1,133 @@
...
+        const isTopLevelOptionChecked = value.some(
```
surchs: now we're going through the values (selected pipelines), but calling them options internally. Then we check which pipeline has the current group as a parent and is also a top-level pipeline. Top level means you selet everything.

src/components/HierarchicalSelectField.tsx
```diff
@@ -0,0 +1,133 @@
...
+
+        return (
+          <CollapsibleSelectGroup
+            key={key}
+            groupKey={key}
+            groupId={group}
+            groupLabel={parentLabel}
+            isGroupChecked={isTopLevelOptionChecked}
+            isGroupIndeterminate={isGroupIndeterminate}
+            groupCheckboxDataCyPrefix={groupCheckboxDataCyPrefix}
```
surchs: 2 different props for different levels of the same state

src/components/CollapsibleSelectGroup.tsx
```diff
@@ -0,0 +1,87 @@
...
+            '&.Mui-expanded': { minHeight: 36, maxHeight: 36 },
+            '&:hover': { backgroundColor: 'action.hover' },
+            '.MuiAccordionSummary-content': {
+              my: 0,
+              alignItems: 'center',
+              '&.Mui-expanded': { my: 0 },
+            },
+          }}
+        >
+          <Checkbox
```
surchs: Group selector

src/utils/utils.ts
```diff
@@ -11,12 +15,109 @@ import {
...
+      return [
+        {
+          id: pId,
+          label: `${pLabel} any version`,
+          parentId: pId,
+          parentLabel: pLabel,
+          isTopLevel: true,
+        },
+      ];
+    }
```
surchs: Check. iiuc this branch should only be hit during the brief interval when pipeline names have been fetched, but the corresponding pipeline versions haven't yet been fetched

src/utils/utils.ts
```diff
@@ -11,12 +15,109 @@ import {
...
+    const pLabel = getPipelineLabel(pId);
+    const versions = pipelines[pId] ?? [];
+    if (versions.length === 0) {
+      return [
+        {
+          id: pId,
+          label: `${pLabel} any version`,
+          parentId: pId,
+          parentLabel: pLabel,
+          isTopLevel: true,
```
surchs: Check: this likely has not actual impact on the rendering vs not rendering / showing of the pipeline name as a top-level option / parent option in the hierarchical view (instead this is done via renderGroup and groupBy

src/App.tsx
```diff
@@ -55,6 +57,9 @@ function App() {
...
+  // Track pipeline IDs for which version fetching has already been initiated to prevent
+  // duplicate in-flight requests or infinite fetch loops when a pipeline has no versions.
+  const fetchedPipelineURIs = useRef<Set<keyof Pipelines>>(new Set());
```
surchs: Check

src/components/HierarchicalSelectField.tsx
```diff
@@ -0,0 +1,133 @@
...
+      // Uncheck parent: remove all selections for this group
+      updatedValue = value.filter((opt) => opt.parentId !== parentId);
+    } else {
+      // Check parent: clear any specific child options and add the top-level group option
+      const filtered = value.filter((opt) => opt.parentId !== parentId);
+      updatedValue = [
+        ...filtered,
+        {
+          id: parentId,
+          label: `${parentLabel} any version`,
```
surchs: Check

src/components/HierarchicalSelectField.tsx
```diff
@@ -0,0 +1,133 @@
...
+        return (
+          <CollapsibleSelectGroup
+            key={key}
+            groupKey={key}
+            groupId={group}
+            groupLabel={parentLabel}
+            isGroupChecked={isTopLevelOptionChecked}
+            isGroupIndeterminate={isGroupIndeterminate}
+            groupCheckboxDataCyPrefix={groupCheckboxDataCyPrefix}
+            onToggleGroup={handleToggleGroup}
```
surchs: check.

surchs (approved):
Thanks @rmanaem. As discussed, let's merge this as is and then we take another look at the implementation together after. Could you please make an issue for this task.

I did some (very basic) commenting to help our memory later.

🧑‍🍳 

And: let's not yet release off of this PR as said in https://github.com/neurobagel/planning/issues/401

